PR #49, opened 2024-12-04
Update Databases category description and add MySQL MCP server

designcomputer:
- Updated Databases category description to better reflect different security approaches
- Added MySQL MCP server implementation with configurable access controls


punkpeye:
Thank you!

PR #60, opened 2024-12-05
Add Docker MCP Server

QuantGeekDev:


QuantGeekDev:
Here is the server: https://github.com/QuantGeekDev/docker-mcp
and demo: https://youtu.be/y9X3jOip8hk

PR #77, opened 2024-12-08
Adding typescript implementation of kubernetes mcp server

Flux159:
Adding a separate kubernetes cluster mcp server written in typescript with slightly different goals like helm chart compatibility in the future.


SecretiveShell:
is there any specific reason this is marked as cloud? As far as I can tell this does not rely on any cloud services (Kubernetes can be ran locally on self hosted hardware)

Flux159:
You can run kubernetes locally or via cloud managed services - I only tested on GKE against my production cluster when developing and the other kubernetes server was also marked with a cloud.

PR #87, opened 2024-12-09
Feature/mcp hfspace. Suggestion as to whether to extract an AI Category for this and ChatGPT?

evalstate:


punkpeye:
Merging as is, but open to suggestions for how to adjust the categorization.

PR #96, opened 2024-12-11
Split out a new category for running commands.

g0t4:
Running commands and scripts makes it possible to do the work of many other MCP servers, with a single tool call. Definitely deserves its own category


punkpeye:
Thank you!

g0t4:
My pleasure, my mind is blown with how useful this stupid simple tool is!

PR #116, opened 2024-12-13
Fixed name of rectalogic/langchain-mcp

isaacwasserman:
Integration was previously labeled as `@modelcontextprotocol/server-langchain`, which does not exist. The correct identifier is `rectalogic/langchain-mcp` (which is an MCP client, not a server).


SecretiveShell (commented):
if this is a client should we add it to the client section or leave it as a framework?

perhaps we need a client vs server framework section @punkpeye

punkpeye:
Let's add a client section. Good call!

isaacwasserman:
It looks like the client section has been added, but this line remains in the frameworks section and still has a misleading name.

PR #138, opened 2024-12-18
Add Postman API

delano:
https://github.com/delano/postman-mcp-server


siderakis:
**Parked until later this week — Nick's call.** The PR (punkpeye/awesome-mcp-servers#13426) stays open but is not progressing.

The blocker is a prerequisite, not the entry itself. Their bot now requires a Glama listing plus a score badge in the line before a new entry is merged. Evidence that this is enforced rather than boilerplate: of the 9 entries merged in the last 14 days, 8 carry a badge. The well-known servers that lack one — github/github-mcp-server, cloudflare/mcp-server-cloudflare, getsentry/sentry-mcp, microsoft/playwright-mcp, upstash/context7, grafana/mcp-grafana — were all merged before the rule, so they are grandfathered, not exceptions we can point at. Across all 3,483 entries the badge rate is 60%.

**When we come back to this, decide Glama first, because #138 is downstream of it.** Once the render-truncation finding landed (this section sits at byte 1.00 MB in a 1.40 MB README and GitHub stops rendering at ~512 KB), the remaining argument for the awesome-list entry was that its raw file feeds Glama's directory. So Glama is the placement with the actual traffic and this PR is the byproduct.

Two coherent outcomes, no middle:

1. Create a Glama listing (needs an account; API key at glama.ai/settings/api-keys), add the badge, update the PR. Gets both placements.
2. Skip Glama and close this too, rather than leaving a PR open against a step we have decided not to take.

PR #155, opened 2024-12-26
add mcp-difyworkflow-server

gotoolkits:
mcp-difyworkflow-server is an mcp server Tools application that implements the query and invocation of Dify workflows, supporting the on-demand operation of multiple custom Dify workflows.


punkpeye (changes requested):
Please limit changes only to adding the server

PR #215, opened 2025-02-14
Add `mcp-server-apple-reminders`

FradSer:
`mcp-server-apple-reminders` is an MCP server for interacting with Apple Reminders on macOS.


punkpeye:
Please limit the PR to adding the server.

PR #336, opened 2025-03-20
Add taskwarrior server

awwaiid:
Also adds a Linux emoji because why not; though then I realized for taskwarrior there is no need to specify since it runs on windows, osx, and linux.


punkpeye:
Thank you. Are you in our Discord?

awwaiid:
I joined the discord now, thanks! :)

PR #355, opened 2025-03-22
Update README: Add multi-database MCP server built with Golang

linhdmn:
Add multi-database MCP server built with Golang, supporting MySQL & PostgreSQL
As an alternative for the [https://github.com/FreePeak/db-mcp-server](https://github.com/FreePeak/db-mcp-server)


punkpeye:
Please use emojis from the index and remove emojis not in the index.

linhdmn:
Hi @punkpeye , I've updated the emojis from the index. Thanks for your time.

PR #408, opened 2025-03-28
Add Arcade.dev

wdawson:
Adding [Arcade.dev](https://www.arcade.dev/) to the Frameworks section.

I used Google Translate for the Chinese and Japanese versions, so I hope they are ok!

### DIFF
--- a/README-ja.md
+++ b/README-ja.md
@@ -256,6 +256,7 @@ Gitリポジトリおよびバージョン管理プラットフォームとの�
 
 ## フレームワーク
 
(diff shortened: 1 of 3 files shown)


punkpeye:
Please create a new section SDKs for this.

wdawson:
Hey @punkpeye . Arcade.dev isn't just an SDK. Arcade provides a service (that you can self-host) to connect any LLM to arbitrary MCP servers. You can use their SDK to build your own tools, but that's not required. 

Is there a better section to put Arcade.dev?

PR #414, opened 2025-03-29
Added DesktopCommanderMCP to Command Line tools.

nbanks:
This is a popular MCP server for executing commands and editing files.  I've been using it on Linux for programming, and I believe it works on MacOS & Windows too.  Since it works by executing commands, even if it's running a test for a program which it's also editing, I think this belongs with CLI tools rather than programming tools.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,6 +128,7 @@ Cloud platform service integration. Enables management and interaction with clou
 
 Run commands, capture output and otherwise interact with shells and command line tools.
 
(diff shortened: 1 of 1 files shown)


punkpeye:
Please align emojis with the legend.

nbanks:
The Desktop Commander MCP is designed to execute things on the command line, read/write/search files, and modify code.  It's written in Typescript and works on the local machine.  What are the appropriate emojis for this?

PR #415, opened 2025-03-29
[add new security mcp server] remote attestation..

maceip:
adds a Security MCP server link demonstrating an MCP server inside an Intel SGX trusted execution environment.

Bonus: Uses RA-TLS to showcase remote attestation.


punkpeye:
Please limit emojis to those in the legend.

PR #516, opened 2025-04-02
Update README.md

non-npc:
added Vibe Model Context Protocol Server to the list


punkpeye:
Please use GitHub repository name as the label of the link.

PR #545, opened 2025-04-04
Add new git and memory server

kwanLeeFrmVi:

PR #598, opened 2025-04-08
Add mcp-rag-server entry to multiple language README files

kwanLeeFrmVi:
split to single pr for each mcp server
ref: #545


punkpeye:
Please remove use of custom emojis

PR #358, opened 2025-03-22
Update README.md to include a new Stata MCP server

hanlulong:
Added new Stata MCP server

PR #660, opened 2025-04-13
Update README.md to include a new Stata MCP server

hanlulong:
The previous change was not make correctly. see https://github.com/punkpeye/awesome-mcp-servers/pull/358


punkpeye:
I don't think this is a server, though?

hanlulong:
It opens a local server to allow users to use Stata with AI models. It is not a public server (as it will violate Stata license agreements).

PR #661, opened 2025-04-13
docs: Update README.md to add a new Ebook MCP

onebirdrocks:
I updated the README file to include a new third-party MCP server named ebook-mcp, which supports reading and interacting with PDF and EPUB files.

Motivation and Context
This MCP Server enables users to interact with PDF and EPUB documents through LLMs. It expands the ecosystem of MCP-compatible tools by offering document-based interaction, which is useful for users working with ebooks or long-form content.


punkpeye:
* Please use standard emojis, like the other servers do
* Please use full repository name for the link label

PR #746, opened 2025-04-23
Update translations

frederico-kluser:


punkpeye:
Looks like your version adds hard breaks to translations. That breaks existing conventions.

Appreciate the effort, but can only merge this if the hard break changes are undone.

frederico-kluser:
Thanks on your review, when i get more time i will try again, and thank for your work on this repo

PR #819, opened 2025-05-01
Added MCP STDIO to Streamable HTTP Adapter

ferrants:
Added MCP STDIO to Streamable HTTP Adapter so Clients that support STDIO can start using some of the Streamable HTTP servers and so Streamable HTTP servers can be developed.

https://github.com/pyroprompts/mcp-stdio-to-streamable-http-adapter


punkpeye:
This is a better addition to https://github.com/punkpeye/awesome-mcp-devtools

PR #824, opened 2025-05-02
Add VideoDB Director MCP

omgate234:
Adding VideoDB Director MCP server - Create AI-powered video workflows including automatic editing, content moderation, voice cloning, highlight generation, and searchable video moments—all accessible via simple APIs and intuitive chat-based interfaces.


punkpeye:
Please add only once and use the full repository name for the link label.

PR #839, opened 2025-05-05
added bedrock mcp connector

VishalYadavCF:


punkpeye:
This is a better fit for https://github.com/punkpeye/awesome-mcp-devtools/

PR #841, opened 2025-05-05
Added a word editing mcp server that allows to replace words in a word file

thibautleaux-kreactive:
ability to edit a word file, clone and list files locally and also transform it to pdf (very usefull to use with a template to fill it with cvs for example)


punkpeye:
Please use the full name: KS-GEN-AI/word-mcp-server

punkpeye:
Please re-open when the comments are addressed.

PR #895, opened 2025-05-16
Remove duplicate metatool-app

technicalpickles:
It was on here twice: once as the GitHub org + repo, once as just the name


punkpeye:
Thank you

PR #1017, opened 2025-06-13
Sorted every entry alphabetically, added "Document Processing" category, relocated entries into more suitable categories, fixed typos, added username entries where it is missing

mikavehns:


punkpeye:
Please don't. It will create thousands of merge conflicts. Can only do this when there are no open PRs.

PR #1036, opened 2025-06-20
Updated the entry for the Prisma MCP server.

gniting:


punkpeye:
We only allow GitHub links.

PR #1212, opened 2025-08-04
Adds a link to keyboard

avb-is-me:
Adds a link to our own MCP server, Keyboard which is a tool that allows anybody to automate any task from their SaaS apps by allowing MCP clients to have dedicated GitHub Codespaces to execute any code related to tasks they might want to automate. The idea is that to integrate a SaaS app you just add the API keys to a protected and private GitHub Codespace(aka only you have access), so that way you don't have to install a 100 MCP apps, but instead just add an API key to this service. The final part of Keyboard is that we have an electron app that allows you approve and deny the code that is about to be executed to prevent things like prompt injection.


punkpeye:
Hey @avb-is-me, closing this since it's been inactive for a while. Feel free to re-open or submit a new PR when you're ready.

PR #1259, opened 2025-08-17
fix sentry mcp link

bruno-garcia:
The current link leads to a 404.

That repo was archived, see: https://github.com/modelcontextprotocol/servers-archived/tree/main/src/sentry


punkpeye:
thank you!

PR #1428, opened 2025-10-17
Add Mcpresso to dev tools

valentinsimplifier:
OAuth 2.1 authentication framework for MCP servers with user management, JWT tokens, and database-backed session management


valentinsimplifier:
@punkpeye Let me know if there is anything else I should add here :)

PR #1518, opened 2025-12-05
Add new MCP server (converting REST API  into MCP tools)

bbonnin:
Add new MCP server that  converts any REST API defined with OpenAPI/Swagger into MCP tools. 
It can be used to enable the adoption of the MCP protocol in context where you have multiple REST API, but no time to provide MCP interface for the API.


tysoncung:
Converting REST APIs to MCP tools is a great use case - it lowers the barrier for integrating existing APIs with AI agents.

Could you add a brief description of what makes this approach different from manual tool definition? That would help users understand the value proposition.

bbonnin:
Thank for your comment @tysoncung ! I have made an error and I will create a new PR, Sorry for that ...

PR #1725, opened 2026-01-23
ci: add weekly automated release workflow

yordis:
## Motivation                                                               

GitHub notifications for PRs on popular repositories like this one can be overwhelming.

By enabling automated weekly releases, users can subscribe to **Release notifications only** and receive a consolidated weekly summary of changes instead of being notified for every individual PR.

This makes it easier to keep up with the project's progress without notification fatigue.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,61 @@
+name: Weekly Release
+
+on:
+  schedule:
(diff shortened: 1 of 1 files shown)


yordis:
@punkpeye it is CI, so take it with a grain of salt, but I would seriously appreciate if we can make this to work 🙏🏻

PR #2636, opened 2026-03-02
Add ckanthony/Chisel to File Systems section

ckanthony:
Rust-powered MCP with patch-based edits and kernel-enforced path confinement for token-efficient, sandboxed file access.


ckanthony:
@punkpeye, I had completed the follow-up, nice work and hope to be get listed.

PR #3284, opened 2026-03-15
Add MCP Synapse to MCP frameworks list

solidsynapse:
Add MCP Synapse to the Frameworks section.

MCP Synapse is a provider-agnostic MCP control plane for managing 
connections, routing requests, and policy enforcement across AI vendors. 
BYOK, local-first, supports OpenAI, Vertex AI, Azure, Ollama, and HuggingFace.

GitHub: https://github.com/solidsynapse/MCP-Synapse
Site: https://mcpsynapse.dev


solidsynapse:
MCP Synapse is a provider-agnostic control plane that manages 
and routes to MCP servers — not an MCP server implementation itself. 
Glama listing requires a Docker image and hosted endpoint which 
doesn't apply to this desktop application. Happy to discuss 
if there's a better way to handle this.

punkpeye:
This repository specifically curates MCP server implementations, so a control plane or routing layer that manages other servers may not be the right fit here. You might want to check out the [awesome-mcp-clients](https://github.com/punkpeye/awesome-mcp-clients) list instead, which may be a better home for tools that manage or route to MCP servers. If you believe MCP Synapse does also function as an MCP server, the Glama badge and AAA scores are still a hard requirement for listing here.

solidsynapse:
> This repository specifically curates MCP server implementations, so a control plane or routing layer that manages other servers may not be the right fit here. You might want to check out the [awesome-mcp-clients](https://github.com/punkpeye/awesome-mcp-clients) list instead, which may be a better home for tools that manage or route to MCP servers. If you believe MCP Synapse does also function as an MCP server, the Glama badge and AAA scores are still a hard requirement for listing here.

Thanks for the clarification. MCP Synapse is better described as a control plane/router than a direct MCP server implementation, so this doesn’t seem like the right list. I’ll close this PR and submit it to a more appropriate list instead.

PR #3422, opened 2026-03-18
docs: add LINE WORKS branding to nworks description

yjcho9317:
Update nworks description to include LINE WORKS branding.
LINE WORKS is the global brand name for NAVER WORKS, used in Japan and Southeast Asia.
This is not a duplicate - only updating the existing description to include LINE WORKS branding.


yjcho9317:
Labels are from the bot's auto-detection. This is not a new entry - only updating the existing description to add LINE WORKS branding. Glama badge is already included in the entry.

PR #2847, opened 2026-03-07
Add bitcoin-mcp to Finance & Fintech

Bortlesboat:
Adds [bitcoin-mcp](https://github.com/Bortlesboat/bitcoin-mcp) to the Finance & Fintech section.

Bitcoin Core MCP server with **43 tools**, 6 prompts, and 7 resources for querying blocks, transactions, fees, mempool, mining, market sentiment, and price data. Fee intelligence that saves money on every Bitcoin transaction. Connects AI agents to a self-hosted Bitcoin node or remote Satoshi API (zero install).

- **PyPI:** https://pypi.org/project/bitcoin-mcp/ (v0.4.0)
- **103 tests**, MIT licensed
- **Docker:** docker run bitcoin-mcp
- **Glama:** https://glama.ai/mcp/servers/@Bortlesboat/bitcoin-mcp

<a href="https://glama.ai/mcp/servers/@Bortlesboat/bitcoin-mcp">
  
</a>

PR #3439, opened 2026-03-18
Update bitcoin-mcp tool count (43 -> 49)

Bortlesboat:
bitcoin-mcp has added 6 new tools since the listing was added: indexed address balance/history, indexed transaction lookup, indexer status, PSBT security analysis, and inscription listing security guide.


Bortlesboat:
Closing — the bot correctly flagged that bitcoin-mcp is already listed (added in #2847). The tool count update isn't significant enough for a separate entry change. Thanks\!

PR #3515, opened 2026-03-19
Add WebdriverIO MCP server

Winify:
Adding Official [webdriverio/mcp](https://github.com/webdriverio/mcp) server to the list.


Winify:
The server shows AAA Score now

PR #4914, opened 2026-04-15
Add jswallez/save-markdown to Knowledge & Memory 🤖🤖🤖

jswallez:
Adds Save to the Knowledge & Memory section.

**What it is:** Save captures any webpage as clean Markdown to a local vault (`~/Documents/Save Vault/`) via a Chrome extension + macOS menu bar app, then exposes the vault to Claude via a built-in MCP server. Claude reads the user's saved pages before answering, so responses are grounded in their own curated knowledge.

**Why Knowledge & Memory:** the MCP server provides persistent, queryable personal knowledge storage that Claude pulls from across sessions — the same slot as Obsidian/Zotero/GistPad integrations already in this section.

**Entry details:**
- 🏠 Local Service (reads files from the local vault)
- 🍎 macOS (it's a macOS menu bar app)
- No language emoji (Swift, which doesn't have one in the legend)

Inserted alphabetically between `jinzcdev/markmap-mcp-server` and `kael-bit/engram-rs`.

Website: https://www.savemarkdown.co


jswallez:
Closing — Save is closed-source so I can't meet the Glama listing requirements (public Dockerfile + introspection). Thanks for maintaining the list!

PR #5805, opened 2026-05-03
Add Copper (client-side MCP tool-call observability CLI) 🤖🤖🤖

noelschwarz:
Adds [Copper](https://github.com/noelschwarz/copper) under **Clients** — a TypeScript CLI that wraps the MCP client to stream tool calls, redact secrets, and flag risky calls. By [Basematter](https://basematter.dev). Not an MCP server; fits the client tooling section alongside the existing client directory links.

Made with [Cursor](https://cursor.com)


noelschwarz:
Thanks for the automated review. After reading the requirements Realized Copper is actually an MCP client-side observability tool rather than a server, so it doesn't fit this list, closing the PR. Apologies for the noise.

PR #6023, opened 2026-05-07
Add giuliohome-org/doc-manager (Knowledge & Memory)

giuliohome:
## What

  [doc-manager](https://github.com/giuliohome-org/doc-manager) — Rust + Rocket app that bundles a document vault on Azure Blob Storage with a built-in MCP Streamable HTTP server.

  ## Why it fits Knowledge & Memory

  - Five tools: `list_documents`, `get_document`, `search_documents`, `create_document`, `update_document` — typical personal-vault surface for "talk to your docs" use cases.
  - Encryption-aware: documents stored with client-side AES-256-GCM (`DMENC1:` envelope, password never leaves the browser) are surfaced to Claude with `encrypted: true` flag but never decrypted server-side. Plaintext documents are fully accessible. Zero-knowledge
  guarantee preserved.
  - Single Rust binary serves both the React UI and the MCP endpoint.


giuliohome:
doc-manager isn't a self-contained MCP server: it's a full document-vault application that exposes an MCP surface, and requires Azure Blob credentials at startup. Happy to add a MCP_PUBLIC_INTROSPECT flag to satisfy the Glama check if that's a hard requirement, but wanted to flag the architecture first.

PR #4166, opened 2026-04-04
🤖🤖🤖 Add QuantOracle — 63 quant computation tools for financial agents

fel123:
## QuantOracle

63 deterministic quant computation tools for autonomous financial agents.

- **Repo**: https://github.com/QuantOracledev/quantoracle
- **API**: https://api.quantoracle.dev
- **Free tier**: 1,000 calls/IP/day, no signup, no API key
- **Categories**: Options, derivatives, risk, portfolio, statistics, crypto/DeFi, macro/FX, TVM
- **Stack**: Python (FastAPI), Cloudflare Worker, x402 micropayments
- **Accuracy**: 120 citation-backed benchmarks (Hull, Wilmott, Bailey & Lopez de Prado)
- **License**: MIT

PR #6091, opened 2026-05-09
Update QuantOracle entry: link to interactive calculators

fel123:
The QuantOracle MCP server (originally added in #4530) recently launched a companion site at [quantoracle.dev](https://quantoracle.dev) with 12 free interactive calculators backed by the same engine — Black-Scholes with Greeks, Monte Carlo, Kelly, VaR, Sharpe, CAGR, crypto liquidation, impermanent loss, and more.

This is useful context for readers browsing the directory: they can verify the tools before setting up the MCP client. Same entry, one-line addition pointing at the live URL.

Diff: +1 line.


fel123:
Thanks for flagging — the duplicate-check bot was correct.

The README had two QuantOracle entries (lines 1455 and 1457) that survived from earlier PRs #4166 and #4530 both merging. I've removed the older stub entry on line 1457 and kept the more polished one on line 1455 (which already has the Glama score badge required by the listing bot, the full description, and the calculator-companion link added in this PR).

Updated diff: now +1 / -1 line. Ready for re-review.

fel123:
Rebased onto current main and squashed to a single one-line diff. This resolves the earlier flags: the duplicate entries are gone from main (one entry remains at its sorted position), the Glama badge is included, and the entry now reflects the server's current surface (79 tools — the original 63 calculators plus composites, batch, live market data, and 24/7 position monitoring). Thanks for the patience — should be a clean review now.

PR #6107, opened 2026-05-09
Fix indentation for kr-crypto-intelligence entry

bakyang2:
## What
Fix incorrect indentation on the kr-crypto-intelligence entry.

## Issue
The line currently starts with `-- [` (double dash) instead of `- [` (single dash), inconsistent with all other entries in the section. This may cause unexpected markdown rendering.

## Fix
Single character change: `--` → `-` at the start of the line.

## Verification
Diff shows minimal change (1 line, indentation only). Content unchanged.


bakyang2:
This is not a new addition — `bakyang2/kr-crypto-intelligence` is already listed at line 1260 (Glama badge included). This PR fixes a single-character indentation typo where the entry incorrectly starts with `--` (double dash) instead of `-` (single dash), making it inconsistent with all other entries in the section.

The diff is a 1-line change with no content modification.

PR #6671, opened 2026-05-20
Add Martin Loop to Coding Agents

Keesan12:
Adds Martin Loop to the developer tools section. Martin Loop is a governed MCP runtime for AI coding agents with budget caps, verifier gates, and inspectable runs.


Keesan12:
Rebased this branch onto current main and refreshed the Martin Loop entry so it points at the top-level repo and the current published MCP package (
px -y @martinloop/mcp).

PR #7160, opened 2026-05-30
Refresh Tuning Engines governed runtime MCP listing

cerebrixos:
Thank you for maintaining this directory. Tuning Engines is already listed, so this PR updates the existing entry instead of adding a duplicate. The refreshed copy reflects the current governed-runtime surface: model, agent, skill, and MCP workflows with policy controls, approvals, traces, usage analytics, and the `npx -y --package tuningengines-cli@latest te mcp serve` install command. The existing Glama score badge and platform markers are preserved.


cerebrixos:
Thanks for flagging the duplicate. I updated the PR to replace the existing stale fine-tuning-only description in place rather than add a second entry. The diff is now a one-line metadata refresh that preserves the existing Glama score badge, and the submission check is passing.

PR #7550, opened 2026-06-07
Add OS Automation (操作系统自动化) section to README-zh.md

tinqiao-oss:
## What

Adds the **OS Automation (操作系统自动化)** category to `README-zh.md`. The English `README.md` has this category, but the Chinese translation was missing it entirely.

This translates the whole category — heading, intro line, and all **4** entries currently in the English OS Automation section — and adds the matching entry to the Chinese table of contents (placed between 监测 / Monitoring and 搜索 / Search, mirroring the English ordering).

## Details

- Each entry keeps the **same repository link, Glama badge, and platform/language icons** as the English README; only the description is translated to Chinese.
- Entries included: `sbuysse/gnome-desktop-mcp`, `dimpagk92/cellar`, `Harusame64/desktop-touch-mcp`, `tinqiao-oss/clawtouch-mcp`.
- TOC anchor `#os-automation` matches the new section's `<a name="os-automation">`.
- No changes to `README.md` or any other language file.

### DIFF
--- a/README-zh.md
+++ b/README-zh.md
@@ -100,6 +100,7 @@
 * 🗺️ - [位置服务](#location-services)
 * 🎯 - [营销](#marketing)
 * 📊 - [监测](#monitoring)
+* 🖥️ - [操作系统自动化](#os-automation)
 * 🔎 - [搜索](#search)
 * 🔒 - [安全](#security)
 * 🏃 - [体育](#sports)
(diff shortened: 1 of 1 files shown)


tinqiao-oss:
Sharing a diagnosis rather than a nudge — this looks like a workflow blind spot for translated READMEs, not something specific to this PR.

This PR only touches `README-zh.md`, porting the existing **OS Automation** section into the Chinese README. All four entries already exist in `README.md` (lines 2867, 2868, 2869, 2872), and three of the four are other people's servers.

`check-glama.yml` reads the base README with a hardcoded path:

```js
// Read existing README to check for duplicates
const readme = fs.readFileSync('README.md', 'utf8');
```

It then filters added lines down to "genuinely new entries (not already in the base README)". For a translation PR that set is empty by construction — every entry is supposed to already be in the English file. Two things follow:

- **duplicate** fires, correctly by the check's own logic but not by intent.
- **missing-glama** fires on the now-empty array, even though all four added lines do carry Glama badges (visible in the diff).

Any PR adding already-listed entries to `README-zh.md`, `README-ar.md`, or another translation will hit the same pair of labels, and there's no edit to the PR content that avoids it short of removing the entries.

No action needed from my side that I can see — just flagging it in case the check is worth teaching about translation files. Equally happy for this to be closed if porting sections into the translated READMEs isn't something you want PRs for; knowing that would be useful either way.

PR #7963, opened 2026-06-12
Add shayaShav/flatten-mcp to Developer Tools

shayaShav:
Adds flatten-mcp under Developer Tools.

flatten-mcp is an MCP server that flattens Claude Code sessions: it moves bulky tool output into a sidecar file and keeps every prompt and event verbatim, so the same conversation resumes at a lower token count instead of being compacted into a lossy summary.

TypeScript, runs locally over stdio, MIT licensed. Also listed in the official MCP Registry as io.github.shayaShav/flatten-mcp.

Entry placed in alphabetical order within the category.


punkpeye:

PR #8061, opened 2026-06-14
Add lilbee to Search & Data Extraction

tobocop2:
Adds **[lilbee](https://github.com/tobocop2/lilbee)** to the Search & Data Extraction section (alphabetical position).

lilbee is a local-first search engine you can talk to. It runs and manages its own local models, or uses your existing Ollama or LM Studio, indexes your files and code, crawls the websites you point it at, and answers over MCP with a citation back to the source. It ships as a single executable or a `pip install`: no Docker, no containers, no separate vector database or model server to run, which makes it easy to deploy. The model manager is one to one with [llama.cpp](https://github.com/ggml-org/llama.cpp) and pulls any GGUF from Hugging Face.

Site: https://lilbee.sh. There is also a free Obsidian plugin for non-terminal users: https://community.obsidian.md/plugins/lilbee. Multi-GPU serving and an opencode integration are coming in https://github.com/tobocop2/lilbee/pull/267.


tobocop2:
@punkpeye, just a friendly nudge. This was merged, but I don't see lilbee at all in the readme so it looks like it was maybe lost during some conflict resolution.

tobocop2:
> @punkpeye, just a friendly nudge. This was merged, but I don't see lilbee at all in the readme so it looks like it was maybe lost during some conflict resolution.

following up @punkpeye

PR #8077, opened 2026-06-15
Add Rapls AI Chatbot (WordPress MCP server) 🤖🤖🤖

rapls:
Adds Rapls AI Chatbot, a WordPress plugin with a built-in MCP server exposing 5 tools (search_knowledge, list_conversations, get_conversation, send_message, get_site_info). Placed under Developer Tools, after rafsilva85, matching existing WordPress MCP entries. Self-hosted, bring-your-own-key.


rapls:
Thanks for the detailed requirements. Rapls AI Chatbot's MCP server runs inside a WordPress plugin (remote HTTP / JSON-RPC over POST), so it can't be packaged as a standalone Dockerfile that starts and responds to introspection on its own — it requires a full WordPress install, DB, and plugin configuration to run. Since the Glama check can't be satisfied for this architecture, I'll close this PR for now. I may revisit via the connectors route once a public hosted demo endpoint is available. Thanks again.

PR #8229, opened 2026-06-17
Add llmtrim to Developer Tools

fkiene:
Adds [llmtrim](https://github.com/fkiene/llmtrim) under **Developer Tools**.

llmtrim is a self-hosted MCP server (also a proxy, CLI, and library) that compresses LLM prompts, tool outputs, and replies to cut token costs. It exposes three tools:

- `llmtrim_compress` — compress a full request body, honoring your `~/.llmtrim` config
- `llmtrim_compress_text` — lossless compression of a single text blob
- `llmtrim_stats` — your savings ledger

Every compression step is re-measured with the provider's tokenizer and reverted if it does not save, and the cached prefix is never rewritten, so it cannot raise your bill. Measured -31% input and -74% output tokens across 112 live A/B cases.

- Rust, self-hosted/local (tagged 🦀 🏠 🍎 🪟 🐧).
- Open source (AGPL-3.0).


fkiene:
Closing for now — llmtrim isn't listed on Glama yet, so the listing requirement can't be met at this time. May resubmit once it's registered.

PR #8620, opened 2026-06-24
Add Verified Burst (verified pay-per-call inference) to Finance & Fintech

ericutility-prog:
Adds **Verified Burst** under **Finance & Fintech**.

An MCP server where an agent buys a verified inference burst over x402: best-of-N on the caller's own key, checked by an independent judge (a different model family), charged a sub-cent fee only when the answer passes — a miss is free. BYOK, hard per-wallet spend cap, USDC settlement on Base mainnet.

- Install: `pip install verified-burst`
- Try live (no wallet/key): https://burst.solcleus.com
- Package: https://pypi.org/project/verified-burst/

Legend: 🐍 Python client · ☁️ Cloud service. Placed alphabetically; follows the entry format.


ericutility-prog:
Done — Verified Burst is now listed on Glama and passing its checks; score badge added after the entry. Thanks for the review! (And good to connect — turns out you're behind Glama too.)

PR #8792, opened 2026-06-26
Add Vestige to Knowledge & Memory

samvallad33:
Adding Vestige under Knowledge & Memory.

Quick context on why it exists: I kept losing days to bugs where the root cause was some quiet change I made a week earlier that looked nothing like the crash it eventually caused. Every memory tool I tried searched for what looked like the error, but a root cause never looks like the bug it creates, so they all missed it.

So Vestige does the opposite. When a failure gets recorded it reaches backward in time and surfaces the earlier memory that actually caused it, matched on a shared entity (same file, env var, service) instead of keyword similarity. That is the one thing single-vector retrieval provably cannot do. It also flags when a new claim contradicts a decision you already stored.

Details:
- Local-first, one Rust binary, no cloud, data stays on your machine
- npm install -g vestige-mcp-server
- 565 stars, v2.2.0, already on the official MCP Registry
- Entry is one line in Knowledge & Memory, formatted per CONTRIBUTING.md

Thanks for maintaining this list.


samvallad33:
Thanks for the review. Done:

- Vestige is listed on Glama at https://glama.ai/mcp/servers/samvallad33/vestige
- Added the Glama score badge after the repo link in the entry

Let me know if anything else is needed.

samvallad33:
Rebased on latest main, so the earlier conflict is resolved. It is now a single clean commit adding one line to Knowledge & Memory, and the PR shows mergeable.

Glama evaluation is complete: score 92 (grade A), 34 tools detected, the score badge resolves. Ready whenever you have a moment. Thanks for maintaining the list.

samvallad33:
Closing this as a duplicate - Vestige was already added in #8602, which is merged. Apologies for the noise, and thanks for maintaining the list.

PR #8915, opened 2026-06-29
Add mcpdrain (stdio deadlock guardian) to Frameworks

patchwright:
Adds [mcpdrain](https://github.com/patchwright/mcpdrain) to the Frameworks section.

It's a small Rust utility that sits in front of any stdio MCP server (`mcpdrain run -- <server>`) and concurrently drains stdin/stdout/stderr, so the server can't deadlock on a full OS pipe buffer — the silent ~64KB stdio hang (e.g. a large `tools/list` response). One line added, follows the existing entry format (🦀 🏠 🍎 🐧, no Glama badge yet). Apache-2.0, on crates.io.

Thanks for maintaining this list.


patchwright:
Thanks for the quick check! mcpdrain is a transparent stdio *proxy/utility* rather than an MCP server, so it can't pass Glama's server-introspection check (it has no tools/resources of its own — it just keeps the wrapped server's stdio from deadlocking). It really belongs in your **awesome-mcp-devtools** list under *Utilities → Proxies and Gateways*, so I'll send it there instead and close this. Thanks for maintaining the list!

PR #8945, opened 2026-06-29
Add reedox/edgar-events-mcp to Finance & Fintech

reedox:
Adds **edgar-events-mcp** to the Finance & Fintech list.

It gives an agent live SEC EDGAR filing events. The standout tool is `get_recent_activist_stakes`, which reads each SC 13D cover page and returns structured rows (holder, target ticker, percent of class, share count) instead of a link to parse. It needs no API key, so it can be wired up and tested in under a minute. Three more tools cover the filtered 13D feed and typed 8-K / S-1 / merger events.

- Repo: https://github.com/reedox/edgar-events-mcp
- Python, stdio, runs locally; installs today with `pip install git+https://github.com/reedox/edgar-events-mcp`
- MIT licensed, free tool works with no signup

Placed alphabetically under Finance & Fintech.


reedox:
Superseded by a submission from the maintaining org (fortifiedai).

PR #9457, opened 2026-07-06
Update infranode: 12 tools covering 67 data types

street1983nk:
Corrects the stale tool count in the infranode entry. The server's tools were consolidated (previously one tool per data type); it now exposes 12 lean read-only tools covering 67 data types, not 38. One-line change.


street1983nk:
This PR doesn't add a new entry, it edits the existing `street1983nk/infranode` line in place. The diff is a single line (one `-`, one `+`): it corrects an outdated tool count. The server's tools were consolidated (previously one tool per data type), so it now exposes 12 lean read-only tools covering 67 data types, not 38.

So it isn't a duplicate, and the Glama score badge is already part of that entry (the server is listed on Glama, remote endpoint at https://mcp.infranode.dev/mcp). Happy to adjust if you'd prefer corrections submitted a different way.

PR #9486, opened 2026-07-06
Fix godot-mcp-runtime (UDP->TCP) and refresh fetch-guard descriptions

Erodenn:
Two small description corrections for servers I author.

**Erodenn/godot-mcp-runtime** — the entry said "injected UDP bridge," but the server now uses a **TCP** bridge (see its [README](https://github.com/Erodenn/godot-mcp-runtime#what-it-does) and source). Corrected the protocol and clarified that it's a zero-footprint server (no addon committed to the project), which distinguishes it from the addon-based Godot runtime servers.

**Erodenn/fetch-guard** — refreshed to reflect current capabilities: the injection scanner now compiles **51 patterns across 7 languages** with a four-phase scan (homoglyph, base64/hex, ROT13 decoding), not "15-pattern," and it added optional JavaScript rendering. The three-layer defense structure is unchanged.

No new entries, no reordering; only the two description strings changed.


Erodenn:
Quick clarification for the maintainer, since the bots flagged this as if it were a new submission:

- **duplicate-check** is a false positive here. This PR doesn't add new servers, it edits the two existing entries in place (`Erodenn/godot-mcp-runtime` at line 1868 and `Erodenn/fetch-guard` at line 2688). The URLs it lists as "already listed" are precisely the entries being edited, so there's nothing to remove.
- **glama-check**: both entries already include their Glama score badges, unchanged in this diff, so no badge needs adding.

The diff is description-only and touches just those two lines: correcting `UDP` to `TCP` on godot-mcp-runtime, and refreshing fetch-guard's capabilities (updated pattern count and added JavaScript rendering). I author both servers.

PR #9519, opened 2026-07-06
Add specshield-mcp-server

specshield26:
Adds SpecShield — the API-change deploy gate for AI coding agents. Read-only / analyze-only.


specshield26 (commented):
Adds SpecShield — the API-change deploy gate for AI coding agents. Read-only / analyze-only."

PR #9656, opened 2026-07-08
Add queue-inspector-mcp

Yusufihsangorgel:
Adds [queue-inspector-mcp](https://github.com/Yusufihsangorgel/queue-inspector-mcp) to the Databases section, next to the existing Redis entries.

It's an MCP server for inspecting and operating Redis-backed job queues — Asynq (Go) and BullMQ (Node): per-state counts, individual job detail (decoded payload, attempts, last error), and retry/delete that run each library's own atomic scripts so the semantics match exactly. A read-only mode omits the mutating tools entirely. TypeScript, MIT, tested against real producers on a real Redis.


punkpeye:
Fixed the entry formatting — moved the badge to the correct position.

PR #9670, opened 2026-07-08
Add mcp-sso — OAuth 2.1 DCR bridge for MCP servers

acartag7:
## mcp-sso

OAuth 2.1 bridge so MCP clients can self-onboard against enterprise identity providers (Entra ID, Cloudflare Access) that don't support Dynamic Client Registration (DCR).

**What it does:** speaks DCR + PKCE + consent to the MCP client; your IdP stays the identity source; mints its own audience-bound tokens (upstream IdP tokens never pass through).

**Category:** Security — it's auth infrastructure for MCP servers, not a server itself. Happy to move if there's a better fit.

**Proof:** live-verified against 5 real MCP clients (Claude Code, Codex CLI, claude.ai, ChatGPT, MCP SDK) on a real Cloudflare Access tenant. Also tested with Entra ID in a real enterprise deployment.

**Stack:** TypeScript, jose-only runtime dep, MIT, npm: https://www.npmjs.com/package/mcp-sso

Repo: https://github.com/acartag7/mcp-sso


acartag7:
Thanks for the update! mcp-sso is actually auth infrastructure (a library/bridge),
not an MCP server itself — it's the OAuth layer you put in front of a server. It
doesn't have a standalone server that responds to introspection on its own, and
even the example requires OAuth auth before /mcp responds (so an automated check
would correctly get a 401).

If there's ever a category for MCP tooling/infrastructure, I'd love to be
included. For now I'll withdraw this PR — no hard feelings. Thanks for curating
the list!

PR #9835, opened 2026-07-11
Fix Ariadne retrieval description

whyy9527:
## Summary

- replace the stale TF-IDF and optional embeddings claim
- describe Ariadne’s current deterministic token scoring implementation

## Verification

- README-only metadata correction
- repository link and list ordering are unchanged


whyy9527:
This PR updates the description of the existing Ariadne entry; it does not add a duplicate entry. The submission check is currently passing.

PR #9920, opened 2026-07-12
Update README with new Apify scraper links

CBLU2005:
Added new Apify scrapers for US building permits, SAM.gov opportunities, and NPI healthcare providers.


CBLU2005:
Thanks for the clear requirements! These tools are hosted services rather than GitHub-hosted servers, so they don't fit the list's current shape — closing for now. May return with an open-source wrapper repo that meets the criteria.

PR #9921, opened 2026-07-12
Omnicord: correct the tool count and add the install command 🤖🤖🤖

OrygnsCode:
Two corrections to the existing Omnicord entry.

**Tool count.** It said 148, which was accurate when it was added. The project is at 151 now, so this changes it to "150+" so the entry does not go stale again on the next release.

**Install command.** The entry had none, so tools that ingest this list have to guess one. At least one directory derives `npx -y <RepoName>` from the repo name, which here produces `npx -y Omnicord`. That is not the package. The real one is scoped, `@orygn/omnicord`, and the bare name resolves to an unrelated package on npm, so anyone following the guessed command installs the wrong thing. Putting the real command in backticks fixes it at the source, because those ingesters use a backticked command when the description provides one.

No other entries touched.


OrygnsCode:
Both bots are misfiring here, so to save a maintainer the trip:

This is not a new submission. It edits the existing Omnicord line in place, so the duplicate checker sees the repo URL on both the removed and the added line and flags it as already listed. The diff is one line changed, nothing added.

The Glama badge is likewise already present, and the new line preserves it unchanged.

PR #10332, opened 2026-07-17
Add truecopy (agent skills & MCP supply-chain gate) to Security

askalf:
Adds askalf/truecopy to the Security section. Supply-chain gate for agent skills and MCP servers: poison-scan, hash-pin into a committed lock, verify drift in CI; truecopy-mcp proxies a live server down to its pinned, unmodified tool set. MIT, zero-dep ESM, on npm (@askalf/truecopy). Also stands a weekly public scan of the official Claude Code plugin directory (badge in the README).


askalf:
Submitted truecopy to Glama (7/19, resubmitted 7/20) — waiting on review. The repo ships a root Dockerfile that wraps the reference server and pins its tool set at build time, so the introspection checks should pass. I'll add the score badge here as soon as the listing is indexed.

PR #10480, opened 2026-07-19
Add SPARDA (sparda-mcp) — MCP server for Express/FastAPI/Next.js with behavior proofs

zakariagharzouli:
Adds SPARDA to the Security section.

SPARDA is a deterministic, offline gate for AI-written code: it catches when an edit removes a guard, exposes a route, or breaks an invariant — no API key, in the agent's edit loop — and also proves deploys and PRs from a compiled behavior graph. Open-core, on npm as `sparda-mcp`, in the official MCP registry.

- Repo: https://github.com/zyx77550/sparda
- Entry added in alphabetical order, following CONTRIBUTING.


zakariagharzouli:
Closing this PR because SPARDA is already listed. Keeping a duplicate entry would be misleading; the existing listing remains the canonical one.

PR #10509, opened 2026-07-20
Add ariadne — affected-test selection for Kotlin/Android

MikhailHal:
Adds [ariadne](https://github.com/MikhailHal/ariadne) under **Developer Tools**.

**What it does:** given a Kotlin/Android project path, it runs `git diff` internally (including uncommitted changes) and returns only the unit tests affected by those changes, so an agent can verify a change without running the whole suite.

**Why it may be useful here:** it does function-level static analysis with the Kotlin Analysis API rather than file- or module-level heuristics — the call graph covers functions, properties (initializers, delegates, accessors), constructors/`init` blocks and callable references, plus automatic multi-module and Android variant source-set discovery.

**Verification:** measured against [Now in Android](https://github.com/android/nowinandroid) (34 modules, ~268 Kotlin files): ~4s per query, and every unit-test class was checked by changing a function in the class it tests — 25/27 valid targets selected. Known limits (framework-invoked callbacks, reflection, DI-started subscriptions) are documented in the README rather than glossed over.

**Install:** `brew install mikhailhal/tap/ariadne`, or the fat JAR from Releases (JDK 21+). Apache-2.0.

Notes on the entry: placed at the end of the Developer Tools list, matching how that section is currently ordered in practice; used ☕ since the legend has no Kotlin marker and the codebase is JVM.


MikhailHal:
Now, the ariadne is under Glama's review.

MikhailHal:
Done — the server is now listed on Glama and the entry has been updated with the score badge:

https://glama.ai/mcp/servers/MikhailHal/ariadne

Thanks for the review!

MikhailHal:
I have joined the server! and my discord name is halunoyo

PR #10596, opened 2026-07-21
Add ZeroDust — sweep an agent's gas balance to exactly zero 🤖🤖🤖

andresdefi:
Adds [ZeroDust](https://github.com/andresdefi/zerodust) under **Finance & Fintech**.

### What it does

Agents accumulate dust as a byproduct of multi-chain work — arbitrage, bridging, testing, deployment all leave small native balances on chains the agent will never touch again. Sending them costs gas, and you can't send *all* of it because you need gas to send. So a remainder is always stranded.

ZeroDust removes the remainder entirely: the user signs an EIP-7702 delegation and an EIP-712 intent, a relayer submits the transaction and pays the gas, and the contract asserts `address(this).balance == 0` before it can succeed. Same-chain or cross-chain.

### Why it's a good fit for agents

Sweeping needs a signing key, which is a prompt-injection surface. So it's **read-only by default** — the sweep tools only register when both `ZERODUST_ALLOW_EXECUTE=true` and `ZERODUST_PRIVATE_KEY` are set, and destinations are restricted to the agent's own address unless an explicit allowlist is provided. An agent talked into sweeping somewhere it shouldn't still can't reach an address the operator never approved.

### Verified

Live mainnet sweep, Optimism → Base, 2026-07-21: source balance to exactly 0, delegation auto-revoked, 99.88% delivered, 23.2s end to end — [`0x19456ea8…`](https://optimistic.etherscan.io/tx/0x19456ea86ed91097847ddad6d6b8cfd6a5240dedac3b100d1a06eb64c86def6c)

Contract is deployed and verified across 26 mainnets at a single CREATE2 address. Also listed in the official MCP registry as `io.github.andresdefi/zerodust`.

`npx @zerodust/mcp-server`


andresdefi:
Done — the server is now listed and scored on [Glama](https://glama.ai/mcp/servers/andresdefi/zerodust), and I've added the score badge to the entry as requested. The Dockerfile lives in `packages/mcp-server/`, and Glama's introspection check passed (server starts read-only and lists its tools over stdio).

punkpeye:
Thank you for the update and for adding the Glama badge. However, the Glama quality score for your server is not yet evaluated — our check shows the server has not been fully scored by Glama. Please ensure the server has a quality score set on Glama (any grade is acceptable). You can check the current status at https://glama.ai/mcp/servers/andresdefi/zerodust. Once the score is evaluated, the badge will reflect it and we can proceed with merging.

andresdefi:
Thanks for the pointer, and you were right that the score was missing. I dug into why.

The server had never been successfully built by Glama, so there was nothing to score. Glama builds with the **repository root** as the Docker context, but the only Dockerfile in the repo lived at `packages/mcp-server/Dockerfile` and was written for a context of that subdirectory. Its first instruction was:

```dockerfile
COPY package.json package-lock.json* ./
```

There is no root `package.json` in that monorepo, so the copy had nothing to match, the image was never produced, and the `initialize`/`tools/list` handshake never ran. The badge was pointing at a profile that had no evaluation behind it.

Fixed and merged in andresdefi/zerodust#7:

- Dockerfile moved to the repo root with repo-root-relative `COPY packages/mcp-server/...` paths, nested one removed so there is a single unambiguous build spec
- root `.dockerignore` added
- `glama.json` added claiming maintainership
- `serverInfo.version` synced `0.1.0` -> `0.2.1` to match `package.json`

I verified the build by replaying every Dockerfile step against a clean `git archive` of the repo root (`npm ci`, `tsc`, `npm prune --omit=dev`) and then running the full stdio handshake, which returns `serverInfo 0.2.1` and 6 tools with clean JSON-RPC on stdout.

I am now going through the Glama claim + release flow so the build actually runs and the quality score gets populated. Happy for this to sit until the badge reflects a real grade - I will comment here again once it does, rather than have you merge something still showing "not tested".

andresdefi:
Correction to my previous comment: the root cause I gave was wrong, and I want to fix it on the record.

I claimed Glama builds the repo Dockerfile with the repository root as the Docker context, and that a misplaced Dockerfile was the blocker. That is not how this works. Glama **generates** its own Dockerfile from the build configuration in the server admin panel (`FROM debian:trixie-slim`, `git clone` of the repo, then `CMD ["mcp-proxy", "--"]`). It never reads a Dockerfile from the repository.

The real blocker is that the build configuration was empty: build steps `[]` and CMD arguments `[]`. So the image cloned the repo and started `mcp-proxy --` with no server command behind it. Nothing was compiled and nothing was started, so the `initialize`/`tools/list` handshake never ran and no score was produced. Consistent with this, the directory API still reports `"tools": []` for the server. My earlier fix would not have changed that outcome.

I am now filling in the build steps and start command and running the build/release so introspection can actually execute. The changes in andresdefi/zerodust#7 are still worth keeping (maintainer claim via `glama.json`, `serverInfo.version` corrected to match `package.json`), they were just not the thing standing in the way.

No need to merge until the badge shows a real grade. I will report back here once it does.

PR #10866, opened 2026-07-24
Withdrawn

Darko893:
Withdrawn.


Darko893:
On the duplicate check: this is not a new listing. It edits the existing `Darko893/mcp-server` line in place so that the description names the project the server belongs to. The link in the changed line is the one already in the README, which is why the check matches. Net diff is one line, 1 insertion and 1 deletion, no entry added.

On the Glama badge: understood, that one is a separate follow-up on our side and not part of this text change.

punkpeye:

PR #10867, opened 2026-07-24
Add agentcents — neutral LLM cost advisor skill

labham-llc:
Adds agentcents to Developer Tools. Local LLM cost proxy: forecasts each call's cost as an honest p50→p90 range, grades its own forecasts, and recommends the cheapest model that clears the task — local included. Consult mode needs no install; live mode uses the pip-installable proxy. Placed alphabetically under Developer Tools.


labham-llc:
Closing — agentcents is a closed-source, pip-distributed server, so it can't meet the Glama Dockerfile requirement. Listing on the official MCP registry instead.

PR #10920, opened 2026-07-25
Add kaeris-mcp (AI localization)

RaiGanja:
Adds kaeris-mcp — an MCP server that localizes an app's string files (JSON, .arb, .po, .strings, Android XML, and more) into 46 languages with placeholder-safe, reproducible output, callable straight from Claude/Cursor. Install: `pip install kaeris-mcp`. Docs: https://kaeris.dev


punkpeye:
Hi again! Thanks for adding the badge — that's great progress.

The final requirement before merging is that Glama evaluates the server and assigns a quality score. The Glama page for your server shows the quality score is still not set. This is different from the maintenance and license grades (which are fine).

Could you please check on the Glama side — you may need to ensure the server was fully submitted with a Dockerfile so the checks can run and produce a quality score? Once you see the score badge filled in, we'll be good to go.

Let me know if you have questions!

RaiGanja:
Thanks for the nudge — it turned out the quality score was blocked by something real on our side, not by a missing Dockerfile.

The repo has had a Dockerfile all along, but the server inside it could not start. MCP SDK 2.0 removed `mcp.server.fastmcp` (FastMCP moved to `mcp.server.mcpserver.MCPServer`), so any fresh build — including whatever the automated checks run — died on the import before serving a single tool. I reproduced it exactly as an indexer would: built the image from this repo and spoke `initialize` to the container, and got `ModuleNotFoundError`.

Fixed in 0.2.6 (published to PyPI, repo updated). The import now falls back to the new location, so both SDK generations work. Verified after the fix, same way:

```
$ docker build -t kaeris-mcp . && docker run -i --rm kaeris-mcp
→ initialize: {"name": "kaeris-i18n", "version": "0.2.6"}
→ tools/list: 7 tools
```

A `tools/call` against the running container also translates a mounted file end to end.

Whenever Glama re-evaluates, it should get a working server now. Happy to trigger anything on my side if a re-run needs it — and no rush on the merge until the score shows up.

PR #10927, opened 2026-07-25
Update bulatko/vk-mcp-server entry

bulatko:
The entry was added when the server had ten tools, and it is flagged ☁️ Cloud Service — that is wrong, it is a local stdio server.

This corrects the flags to 📇 🏠 🍎 🪟 🐧 and refreshes the description: nineteen tools now, with MCP tool annotations so clients can tell reads from writes, output schemas on every tool, prompts, and a one-click .mcpb bundle.

One line changed, nothing else touched.


bulatko:
Both boxes are already ticked here:

1. The server is listed on Glama — https://glama.ai/mcp/servers/bulatko/vk-mcp-server — with a Dockerfile checked into the repository, and it introspects (tools, prompts and schemas are shown there).
2. The score badge is in the entry, and was before this PR: `[vk-mcp-server MCP server](https://glama.ai/mcp/servers/bulatko/vk-mcp-server)`

This PR only corrects the existing line: the tool count (10 → 19), the description, and the platform icons — it was marked ☁️ (cloud) when it is a local stdio server that runs on macOS, Windows and Linux.

One thing worth flagging for the listing check: until today the server exited when no token was set in its environment, which is exactly the condition an introspection sandbox starts it in. 0.4.3 (released today) starts and serves `tools/list` with no credentials at all; a tool call without a token returns an explanatory error instead. So a fresh Glama build should now see all 19 tools rather than the older snapshot.

Happy to make any change you want to the wording.

PR #8799, opened 2026-06-27
Add john-broadway/proximo to Cloud Platforms 🤖🤖🤖

john-broadway:
Adds **Proximo** — an open-source (Apache-2.0) MCP + A2A server for Proxmox (VE, Backup Server, Mail Gateway).

Distinct from the existing `mcp-proxmox` entry: Proximo's focus is a *trust layer* for AI-driven infrastructure — dry-run plans before any mutation, a hash-chained tamper-evident audit ledger, and undo/snapshot-before-action. 325 tools across the three surfaces.

Live on PyPI (`proximo-proxmox`), GitHub, and GHCR (signed multi-arch image). Placed alphabetically in Cloud Platforms.

PR #10969, opened 2026-07-26
proximo: 365 -> 900 tools

john-broadway:
The `john-broadway/proximo` entry has been accurate since it merged (#8799) apart from the tool count, which was true at v0.21.1.

Proximo is now at v0.25.0 with **900 tools**. Verified against the live MCP registry rather than a doc.

One line changed; nothing else in the entry is touched (it already names all four Proxmox surfaces).


john-broadway:
closing this one. #11720 replaces it with the current count. 906, not 900. same line, same file. sorry for the double.

PR #11020, opened 2026-07-27
Add acikpoz (Turkish construction unit-price catalogs) to Finance & Fintech

gulmezeren2-byte:
Adds **acikpoz** to *Finance & Fintech*, next to my `ihalent` entry (both are Turkish public-procurement data tools).

It turns official Turkish construction unit-price (birim fiyat) catalog PDFs into structured, machine-readable line items so an agent can compute with real cost data instead of guessing at it.

The parsing is coordinate-based, not ML: the price column's x-position is read from the page header, which matters because some sections carry two numeric columns (the real unit price and a separate assembly fee) and a fixed threshold silently reads the wrong one. It also does year-over-year diffing (price moves with Δ and %Δ, added/removed items), deterministic validation, and reports a coverage grade on every parse — category headers and genuine price gaps are surfaced rather than filled in.

Fully offline, no API keys, no network. MIT, tested, on PyPI (`pip install acikpoz`, MCP server via `acikpoz-mcp`) and in the official MCP Registry. Glama score badge included.


punkpeye:
Thanks for following up — I see you've claimed the server on Glama. However, the quality score still hasn't been evaluated yet (showing as `?`). Per our contribution guidelines, the server needs to have its quality score set on Glama before I can merge. Could you please ensure that the evaluation is triggered on the Glama dashboard? If you need any help, feel free to ask.

PR #8378, opened 2026-06-19
Add cammac-creator/ibanforge-mcp to Finance & Fintech 🤖🤖🤖

cammac-creator:
Adds **IBANforge** to Finance & Fintech.

IBANforge is an MCP-native API for **pre-payout counterparty screening**: IBAN validation, OFAC/EU/UN sanctions screening, Swiss BC-Nummer clearing, SEPA/VoP reachability and risk scoring. Official server (I'm the author), TypeScript, cloud-hosted. Free tier (200 req/month) + x402 USDC pay-per-call.

- Repo: https://github.com/cammac-creator/ibanforge
- npm: https://www.npmjs.com/package/ibanforge-mcp
- MCP Registry: `io.github.cammac-creator/ibanforge`

Inserted alphabetically by slug between `cahthuranag` and `carsol`.

PR #11146, opened 2026-07-29
Fix IBANforge entry: correct sanctions sources, add Swiss clearing detail 🤖🤖🤖

cammac-creator:
Corrects the IBANforge entry added in #8378.

**The problem.** The entry claims `OFAC/EU/UN sanctions screening`. IBANforge does not consume any UN list. Its live sources, returned in `meta.sources` on every compliance response, are OFAC, EU, FATF and EPC. The UN claim is simply wrong and I would rather not leave an inaccurate compliance claim in a widely-read list.

**Also improved while editing the same line:** the BIC lookup scale (121,610 entries) and the Swiss QR-IID clearing capability from SIX BankMaster, both of which the previous wording omitted.

One line changed, no other entries touched.


cammac-creator:
Both bot checks look like false positives on this PR, so flagging for a maintainer.

**duplicate-check** reports `cammac-creator/ibanforge` as already listed. It is, and that is the point: this PR edits that one existing line rather than adding a new one. The diff is a single line changed, one removal and one addition at the same position. There is no duplicate to remove.

**glama-check** asks for a Glama score badge after the description. The badge is already present, and was present before this PR too. It is in both sides of the diff:

```
[![cammac-creator/ibanforge MCP server](https://glama.ai/mcp/servers/cammac-creator/ibanforge/badges/score.svg)](https://glama.ai/mcp/servers/cammac-creator/ibanforge)
```

The server is listed at https://glama.ai/mcp/servers/cammac-creator/ibanforge and the badge resolves.

The change itself corrects the description: the previous wording implied full OFAC/EU/UN sanctions screening, when the check is at bank level against the institution BIC rather than against the account holder. Also swapped a frozen entry count for `121k+`, since the reference database is refreshed monthly.

PR #11214, opened 2026-07-30
Add Quant Data (Finance & Fintech) 🤖🤖🤖

celineycn:
Adds **Quant Data** to Finance & Fintech.

- Remote MCP (streamable HTTP) at `https://api.quantdata.uk/mcp` — no key or signup to try; every tool answers 10 calls/day keyless with the same complete response a subscriber gets.
- On the official MCP registry as `uk.quantdata/quantdata`.
- Four tools: day-type probabilities, Weis volume-wave events, options max pain, estimated dealer gamma (GEX).
- The distinguishing thing is measurement honesty: every published percentage traces to a stated test (e.g. pre-registered Weis event win rates on 16 years of ES 5-minute data), failure cases are published alongside, and nothing in any tool output is framed as a buy/sell recommendation. Methodology: https://quantdata.uk/methodology

Verified working against `@modelcontextprotocol/inspector` (tools/list and tools/call, keyless).


celineycn:
Badge added — thanks for the pointer.

One note on the Glama requirement, since this server is a slightly different shape: **it is a remote server, not a locally-installed one.** It is already indexed on Glama at https://glama.ai/mcp/servers/celineycn/quantdata-plugin, and the hosted endpoint is also in Glama's connectors index (submitting it there returns "a server with this URL already exists").

That means the Dockerfile step does not really apply here — there is nothing to build and install locally. The endpoint is a public HTTPS URL:

```
https://api.quantdata.uk/mcp
```

Introspection needs no credentials at all, so it is directly verifiable:

```bash
curl -s https://api.quantdata.uk/mcp -H 'Content-Type: application/json' \
  -d '{"jsonrpc":"2.0","id":1,"method":"tools/list"}'
```

That returns the five tools with full schemas and annotations. It is also live in the official MCP registry as `uk.quantdata/quantdata`.

Happy to do anything else that helps the check pass — just let me know if a remote entry needs something different from the Dockerfile route.

PR #11234, opened 2026-07-31
Add inth3shadows/runecho to Developer Tools 🤖🤖🤖

inth3shadows:
Adds one entry to Developer Tools. Single line, no other changes.

**[runecho](https://github.com/inth3shadows/runecho)** — 🏎️ Go, 🏠 local, 🍎 🪟 🐧.

A deterministic symbol oracle over a locally built index of the repository: whether a symbol exists, what a file declares, and what changed structurally since the last snapshot. It also ships a PreToolUse guard that stops an Edit/Write referencing a function, constant or type the repo does not declare, before the write lands rather than after the build fails.

No LLM, no API keys, no network, no language server — every answer is a parse and a lookup, so the same code produces the same answer on every run and every machine. Parsers for Go, Python, JS/TS, Rust, Ruby and shell. MIT, CI green, releases tagged.

Legend rationale: 🏠 local because it reads the checkout on disk and never calls a remote API.


punkpeye:
This pull request has been closed due to 38 days of inactivity. The entry looks great — all validation checks pass and the Glama configuration is complete. If you'd like to continue, please feel free to reopen or submit a new PR when you're ready. Thanks for your contribution!

PR #11235, opened 2026-07-31
Add inth3shadows/terse to Aggregators 🤖🤖🤖

inth3shadows:
Adds one entry to Aggregators. Single line, no other changes.

**[terse](https://github.com/inth3shadows/terse)** — 🐍 Python, 🏠 local, 🍎 🪟 🐧, MIT, on PyPI as `terse-mcp`.

A transparent proxy that sits in front of any other MCP server and shrinks its tool output before it reaches the model. Lossless by default: the transformed bytes are a denser encoding of the same data, with a round-trip gate asserting an exact inverse over the corpus, so nothing is dropped, summarised, or offloaded to a cache that expires. Repeated calls to the same tool emit a delta against the prior result rather than the whole payload. Per-tool opt-in, no decode step, no model in the loop.

Section rationale: filed under Aggregators because it wraps other MCP servers and everything flows through the one process. Happy to move it to Developer Tools if you would rather group it by purpose than by shape.


punkpeye:
This pull request has been open for 38 days with no activity from the author since it was opened. Per our contribution guidelines, PRs with no author activity for 30 days are closed. Please feel free to open a new PR when you're ready to continue. Thank you!

PR #11264, opened 2026-07-31
Add AccessNexaaly/nexaaly-a11y to Developer Tools

AccessNexaaly:
Adds an MCP server for web accessibility auditing.

I checked the list first: the Browser Automation section has plenty of servers that *read* the accessibility tree in order to drive or snapshot a UI, but I could not find one that audits a page against WCAG, so I put this in Developer Tools instead. Happy to move it if you disagree.

What it does:

- loads the page in a real browser, so it sees computed style and event handlers rather than static markup
- runs axe-core, then adds the checks axe cannot make from the accessibility tree alone. The main one is controls that are styled and wired as buttons but are not focusable, so a mouse user can click them and a keyboard user cannot reach them at all. In the tree those are just divs with text, which is why rule engines do not flag them
- returns the specific code change for each finding, not just a rule id

MIT. Works over stdio, and there is a hosted endpoint at `https://nexaaly.com/mcp` that needs no install.

One line added at the top of Developer Tools, matching the section's newest-first ordering. Legend flags are 📇 (JavaScript) ☁️ (hosted) 🏠 (local).


AccessNexaaly:
Thanks, and understood on wanting only working servers listed.

This one is already on Glama, as a connector rather than a repo entry, because it ships a hosted endpoint:

**https://glama.ai/mcp/connectors/com.nexaaly/a11y-check**

Glama currently reports it as **Healthy**, last tested 2026-07-31 14:33, transport Streamable HTTP. That is their own introspection against `https://nexaaly.com/mcp`, which is the check you are asking for.

Two things I could not do, so I would rather ask than guess:

1. The score badge you linked is on the `/mcp/servers/OWNER/REPO/` path, and `/mcp/servers/AccessNexaaly/nexaaly-a11y/badges/score.svg` currently 404s because the entry lives under `/mcp/connectors/` instead. Is there an equivalent badge for connectors, or would you prefer I get the GitHub repo indexed as a separate server entry as well?

2. I tried to sign up to add the repo directly and Glama's `complete-profile` step fails with "Something went wrong" after GitHub OAuth, so I could not submit it myself yet.

Happy to add whatever badge you point me at, or to close this and reopen once the repo entry exists. Your P.S. suggested connectors for hosted servers, which is why I went that route first.

punkpeye:

PR #11328, opened 2026-08-01
Add Kaidn — fraud investigation with evidence 🤖🤖🤖

bowsetter:
Adds `@kaidn/mcp` under **Security**.

[Kaidn](https://kaidn.io) is a rules-first fraud-scoring API. The engine's founding constraint is that rules decide and the model only narrates, so every verdict carries the raw evidence behind it — which is what makes it useful to an agent rather than decorative. `explain_event` returns each rule that fired with its numbers, so the model can explain *why* a signup was blocked instead of restating a score.

One tool worth calling out: `check_email` returns `canonical`, an identity key that collapses `+tags`, gmail dot tricks and `googlemail.com` onto a single mailbox — so an agent can tell whether two signups are actually the same person.

- Published: `npx -y @kaidn/mcp` (MIT)
- **Read-only by default.** The two mutating tools require `--allow-writes`; config-editing and GDPR-erasure tools are withheld in every mode, because those belong in a UI where a human can see what they are about to do.
- Quota ceiling per process, so an agent in a loop cannot silently burn a customer's monthly plan.
- stdio and Streamable HTTP transports; 16 tests, MIT licensed.

Entry follows the existing Security format (🎖️ official, 📇 TypeScript, ☁️ cloud). One line added.


punkpeye:
Closing this PR due to inactivity. The author has not been active for 42 days. Please feel free to re-open or submit a new PR when ready.

PR #11329, opened 2026-08-01
Add Aethra — self-hosted deploy platform with an MCP control plane 🤖🤖🤖

Authoritt:
Adds one entry to **Cloud Platforms**, alphabetically between `AIops-tools` and `Focus-GTS`.

**[Authoritt/Aethra](https://github.com/Authoritt/Aethra)** — Apache-2.0, .NET 10, self-hosted.

It is a deploy platform (git→Docker builds, reverse proxy, Let's Encrypt, uptime monitoring, VM and per-container metrics) whose control plane is an MCP server rather than a REST API with an MCP wrapper bolted on afterwards. Three things that seemed worth having in the description:

- Deploy tools return the **healthcheck outcome**, not a 201, so an agent cannot truthfully report success for a deploy that did not happen.
- Every tool response carries `next_actions: [{ tool, why, suggested_args }]`, so the agent is not reconstructing the domain model from documentation.
- The agent's scoped API key can deploy to production but **cannot mint API keys or read secrets** — those endpoints are cookie-only.

Flags follow the legend: 🎖️ official, #️⃣ C# codebase, ☁️/🏠 (self-hosted central managing remote machines, same as the existing Portainer entry), 🐧 Linux.

Install is `docker compose up -d --build` from `deploy/`; the README also covers running from source.

---

Submitted by an automated agent, opting into the fast-track process per CONTRIBUTING.md. The diff is a single line and the links are verified reachable.


Authoritt:
Thanks — I went and tested against the Glama requirement rather than guessing, and I do not think this server can satisfy it. Reporting the result so a maintainer can decide, not asking for an exception.

Two independent blockers, both structural rather than fixable:

**1. It cannot start from a Dockerfile alone.** Aethra is a deploy platform with an embedded MCP server, not a standalone server. Started without its Postgres, the process dies in the reverse-proxy config provider, which loads routes in its constructor. That is deliberate: a proxy that boots with zero routes would serve nothing to every host it fronts, so failing closed is the correct behaviour. The container needs a database to reach a running state at all.

**2. The MCP endpoint requires authentication.** `/mcp` is mapped with `RequireAuthorization()` — cookie or scoped API key. An anonymous `tools/list` gets a 401. I could remove that to pass an introspection check, but this server can trigger production deployments and provision databases, so an unauthenticated MCP endpoint would be a real hole traded for a badge. I would rather not be listed than do that.

So the Glama check is measuring something reasonable that this project structurally is not: a server you can `docker run` and introspect. Every published server I looked at in the Cloud Platforms section is a client-side wrapper over a remote API. This one is the platform.

Entirely your call whether an entry like that belongs in the list. If the answer is that the Glama listing is a hard requirement, close this and no hard feelings — the requirement makes sense for the servers it was written for. If a self-hosted platform exposing MCP is in scope, the entry is one line and the links are live.

(Still an agent submission, per the earlier disclosure.)

PR #11341, opened 2026-08-01
Add mcpaudit to Security 🤖🤖🤖

AndrewXuTurtle:
Adds [mcpaudit](https://github.com/AndrewXuTurtle/mcpaudit) to the Security section.

Marked `🤖🤖🤖` per CONTRIBUTING — this PR was prepared by an automated agent working on behalf of the repo owner.

**How it differs from what is already listed there.** The existing security entries mostly scan a server or package *before* you install it. mcpaudit audits the ones you are **already running** — it discovers configs across Claude Desktop, Claude Code, Cursor, Windsurf and VS Code and reports on each configured server. Those are complementary; most people have five or six servers already installed and have never reviewed any of them.

It also analyses **provenance**, not just content. While building it I found a live impersonation package on npm: `@modelcontextprotoco1/server-filesystem` (digit one, not lowercase L), which is a **byte-identical copy of the official `@modelcontextprotocol/server-filesystem` release 2026.1.14** — every file in `dist/` shares the same SHA-256 — republished under a homoglyph scope with forged `author: "Anthropic, PBC"` metadata. There is no malicious code in it, which is the point: content scanners return clean because the content *is* the official implementation. Only provenance catches it. ([writeup](https://github.com/AndrewXuTurtle/mcpaudit/blob/main/ADVISORY.md) — it turned out to be an independent rediscovery; Microsoft Defender ships a signature for the same string.)

**Tools:** `audit_mcp_configs`, `check_package`, `check_advisories`.

**Notes**
- Zero runtime dependencies, including a hand-written tar reader. Taking on a dependency tree in order to report on dependency trees seemed like the wrong trade.
- Nothing is installed and no server is started — `--deep` reads package tarballs in memory.
- Free, no API key, no payment path.
- The protocol is implemented directly against JSON-RPC rather than via the SDK, for the same zero-dependency reason. Verified against the exact command in the entry, from a clean `npx`, before opening this.
- Tuned for precision: findings carry evidence and a confidence tier, and I publish the tool's own [false positives and how they were fixed](https://github.com/AndrewXuTurtle/mcpaudit/blob/main/RESEARCH.md) — five classes so far, including two that would have wrongly flagged innocent maintainers.

MIT. Entry follows the section's existing format; markers are 📇 🏠 🍎 🪟 🐧.


AndrewXuTurtle:
Thanks — the bot is right, and I have removed the badge rather than leave it broken.

The badge URL currently returns **404**: the server has not been evaluated by Glama yet, so it would have rendered as a broken image in the list. Around 1,376 of the entries here carry no badge, so a plain entry seemed better than a broken one.

I have added a [`glama.json`](https://github.com/AndrewXuTurtle/mcpaudit/blob/main/glama.json) at the repo root, which is the discovery convention the indexed servers use. If Glama picks it up and issues a score, I am happy to open a follow-up adding the badge properly.

The entry itself is unchanged otherwise, and I verified the documented command works from a clean `npx` before opening this PR.

punkpeye:
Hi AndrewXuTurtle,

Thank you for your submission. Unfortunately, this PR has been inactive for over 30 days without the required Glama setup being completed. The server needs to be claimed on Glama and have its quality score evaluated before the entry can be added.

Feel free to reopen or submit a new PR once the necessary steps are completed. We'd be happy to review it then.

PR #11352, opened 2026-08-02
Update README with new MCP server entries

nikhilkatyan:


nikhilkatyan (commented):
OptionContext MCP server added for NSE & BSE data

PR #11354, opened 2026-08-02
Add postlint-mcp, apple-podcasts-mcp, and gsc-mcp 🤖🤖🤖

conorbronsdon:
Three servers, all into **existing** categories, alphabetical within each. No new section, so the multi-entry rule doesn't apply.

**🌐 Social Media** — `postlint-mcp`
Checks a post against a platform's real character limit before publishing. The counting is the point: X bills every URL at exactly 23 through t.co regardless of length and weights CJK/Hangul/emoji at 2; Bluesky counts extended grapheme clusters, so a ZWJ family emoji is 1 rather than 11, and carries a second 3,000-byte cap that binds first on emoji-heavy text; Mastodon counts graphemes and charges nothing for the domain on a remote mention. No credentials, no network.

**🎙️ Podcasts** — `apple-podcasts-mcp`
Owner-side Apple Podcasts analytics via the Podcasts Connect Reporter protocol. Distinct from the existing Apple entries in the wider ecosystem, which wrap the public iTunes catalog — Apple's hosting API excludes owner analytics entirely, so plays, listeners, and followers only come over Reporter. Read-only.

**🎯 Marketing** — `gsc-mcp`
Google Search Console for agents. Note the section already has `acamolese/google-search-console-mcp`; this one differs in the striking-distance report (queries at position 8–25 with impressions behind them), the sitemap write tools, and per-tool MCP read/write annotations.

---

Glama badges omitted deliberately — all three badge URLs currently 404, since Glama hasn't ingested these yet. Happy to add them in a follow-up once they resolve, or if you'd rather I hold the PR until then.

The `🤖🤖🤖` suffix is per CONTRIBUTING's opt-in: an agent drafted this, and the disclosure seemed more honest than omitting it. Entries and package names were verified against npm and the live repos.


conorbronsdon:
Closing in favour of one PR per server, which is the pattern in your recent merges — I bundled three here and shouldn't have.

Split into:
- #11355 — postlint-mcp → Social Media
- the Marketing entry for gsc-mcp, opened alongside
- apple-podcasts-mcp → Podcasts, held until its Glama badge resolves (submitted, still in review)

Both open PRs now carry the Glama score badge your bot asked for. Sorry for the noise.

PR #11455, opened 2026-08-03
Add Aura UI server to Developer Tools

BlueStarSystem:
Adds the Aura UI MCP server: a hosted, read-only server for the Aura UI Blade component library (Laravel + Livewire + Tailwind CSS 4). It lets an assistant list and search the components, read a component's real Blade tag, props and documentation, and get the exact install command — so it stops guessing component APIs.

Published on the official MCP registry as io.github.BlueStarSystem/aura-ui. Remote server, nothing to install: https://aura-ui.com/mcp/aura


BlueStarSystem:
Thanks! One clarification: Aura UI is a remote hosted server, not a packaged one — it runs as an endpoint at https://aura-ui.com/mcp/aura, so there is no Dockerfile to add.

Following your P.S., it belongs under connectors. It is already published on the official MCP registry as io.github.BlueStarSystem/aura-ui, which is the same namespace Glama uses to key connectors, so it should be importable without a Docker build. I will add the score badge as soon as the listing exists — happy to adjust if you would prefer a different badge path for hosted connectors.

punkpeye:
Hey @BlueStarSystem, closing this since it's been inactive for a while. Feel free to re-open or submit a new PR when you're ready.

PR #11482, opened 2026-08-04
Add report-workflow (deterministic claim-fidelity gate)

0Smallcat0:
Adds `report-workflow`, a stdio MCP server that verifies claims against the evidence they cite — deterministically, with no model call and no API key, so the same input gives the same verdict every run.

`verify_claims` checks that the numbers, quotes and citations in a draft actually appear in the source it points at, and reports which gate blocked what. Measured rather than asserted: 86.4% of hallucinated claims blocked on a 69-case hand-audited red-team corpus with 0 false positives, and a 0.06% false-positive rate over 10,000 external HaluEval pairs. Both are reproducible from the repo with `--check`.

Scope is stated honestly in the README: it is a fidelity gate, not a hallucination detector — it has no semantics, so a fluent paraphrase that reverses the source passes. The documented misses are kept in the benchmark corpus.

Placed at the top of Developer Tools, matching where recent entries in that section sit; happy to move it if you would rather it went alphabetically.


punkpeye:
This pull request has been closed due to 30 days of inactivity from the author. If you'd like to continue with this submission, please feel free to re-open the PR or submit a new one when you're ready.

PR #11512, opened 2026-08-04
Add dmarc-examiner/mcp server 🤖🤖🤖

lmmartinb:
Adds the official MCP server for DMARC Examiner under 🔒 Security.

**What it does:** exposes DMARC monitoring data for domains you own — aggregate reports broken down by sending source, SPF/DKIM alignment results, security alerts, and CSV export. 8 tools, OAuth-scoped.

**How it differs from what is already listed:** the existing DMARC-related servers (`intodns-mcp-server`, `dechonet-mcp`, `checker-mcp`, `mcp-domain-deliverability-checker`) check a domain's DNS configuration — they read the SPF/DKIM/DMARC records you published. This one reads the aggregate reports that receivers actually sent back, which is a different dataset and requires an account on the service.

- Repo: https://github.com/dmarc-examiner/mcp
- License: MIT
- TypeScript, cloud service, free tier available

Placed at the end of the Security section, matching the surrounding format. Disclosure: I maintain this server, and per CONTRIBUTING the title carries the agent marker.


lmmartinb:
Badge added — the server was already indexed on Glama at [`dmarc-examiner/mcp`](https://glama.ai/mcp/servers/dmarc-examiner/mcp) (id `grtoul4msn`), so the score badge now sits after the description as requested.

On the checks, I want to be upfront rather than game them: **this is a remote server behind OAuth**, and I do not think it can pass the Docker-based quality checks as they stand.

The endpoint is `https://mcp.dmarc-examiner.com/mcp`. Every tool call requires an authorized session, so a container started without credentials and without a browser cannot complete the authorization handshake and will not answer introspection. Adding a Dockerfile would produce a build that starts and then blocks waiting for an authorization that nobody can grant — it would pass the letter of the check while telling you nothing about whether the server works.

What I can say instead, verified today: the npm package (`@dmarc-examiner/mcp`) bridges stdio to the remote endpoint, and against a real MCP client `initialize` succeeds, `tools/list` returns 21 tools and `tools/call` returns live data. It is also MIT licensed — Glama currently reports `license - not found` because it last scanned the repository before the LICENSE file was added this afternoon; that should correct itself on the next crawl.

Your bot also notes that servers with a hosted endpoint can be listed under [glama.ai/mcp/connectors](https://glama.ai/mcp/connectors), which looks like the right home for this one.

Happy to follow whichever route you prefer — list it as is, wait for Glama to re-crawl so the licence grade updates, or close this if authenticated remote servers are out of scope for the list. Disclosure, as in the original PR: I maintain this server.

PR #11521, opened 2026-08-04
Refresh anchor-x402-mcp entry (14 tools, current pricing)

hypeprinter007-stack:
Not a new entry — `anchor-x402-mcp` is already listed, and the duplicate-check bot correctly flagged my original add. Converted this PR to a **refresh of the existing line** instead, so there is no net duplicate (the diff modifies the existing entry in place: −2/+1).

Fixes stale content on the existing entry:
- `16` → **14 stdio tools** (the server exposes 9 commodity + 5 LLM tools).
- "OFAC sanctions screen" → **wallet risk screen** (OFAC sanctions + address-reputation → allow/review/block).
- Removed the async investigator (`$7.77` → it is `$1.77` now anyway) and the signed RNG from the tool list — both are HTTP-only, **not** exposed as stdio tools; noted parenthetically instead.

No change to the Glama badge or placement.


hypeprinter007-stack:
Thanks — the duplicate-check bot is right, `anchor-x402-mcp` is already listed. Ive converted this from an add into a **refresh of the existing entry** (diff is now −2/+1, modifying the existing line in place — no net duplicate), which also corrects some stale details on it (tool count 16→14, OFAC-only→wallet-risk screen, and dropping the HTTP-only investigator/RNG from the stdio tool list). The Glama listing exists and re-crawls the just-published npm `0.2.2`.

PR #11526, opened 2026-08-04
Add effectfence — causal effect fence for agent tool calls (Rust)

aurumflux20:
Adds [effectfence](https://github.com/aurumflux20/effectfence) to Developer Tools.

**What it is:** an MCP stdio server (+ Rust library) that stops the same side effect from executing twice in agent systems — double-charges, duplicate sends, repeated provisioning. Same-instant races elect exactly one winner via a lock-free reservation; late duplicates get a sealed, content-addressed receipt replayed instead of re-running (`fence_prepare` / `fence_commit` / `fence_abort`).

**Signals:** published on [crates.io](https://crates.io/crates/effectfence) · [docs.rs](https://docs.rs/effectfence) · CI runs the test suite plus a 1,000-attempt storm demo that must elect exactly one execution on every commit · MIT.

Entry follows the section format (Rust 🦀, local 🏠, macOS/Windows/Linux).


aurumflux20:
Thanks for the automated check — good catch. effectfence hasn't been evaluated by Glama yet, so that badge would have rendered as a broken image. I've removed it and left a plain entry (matching the ~1,300 other entries without a badge). Happy to add it later once a score exists.

aurumflux20:
Status on the updated requirements:

**Done:** added a `Dockerfile` to the repo, and a CI job that builds the image, boots it, and asserts it answers an MCP `initialize` request — so it should pass the Glama check on arrival ([workflow](https://github.com/aurumflux20/effectfence/actions), all jobs green).

**Pending:** the Glama submission itself — the account belongs to the repo owner, so that step is with them and should be done shortly.

**Deliberately not yet done:** re-adding the score badge. Until Glama has evaluated the server the badge URL 404s, which is exactly the broken-image case your earlier check warns about. I'll add it in the same format you specified the moment a score exists, and comment here when it's up.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

aurumflux20:
Update: effectfence is now listed and evaluated on Glama — https://glama.ai/mcp/servers/aurumflux20/effectfence (score badge resolves). I've re-added the badge in the format the check specified. Should satisfy the listing requirements now — thanks for the patience.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

aurumflux20:
Small update rather than a nudge: effectfence now ships prebuilt binaries for macOS (Apple silicon), Windows x64 and Linux (x64 + arm64), so the entry offers `npx -y effectfence` instead of requiring a Rust toolchain — matching how the neighbouring entries read.

Each binary is built on its own native runner and has to answer an MCP `initialize` on that platform before it's attached to a release; nothing is cross-compiled. It's also listed in the official MCP Registry with both the npm and cargo install paths.

Glama listing and score are in place from the earlier round, so I believe the listing requirements are all met. Happy to adjust anything if something's still outstanding — and no rush, I know the queue is long.

aurumflux20:
Friendly status update rather than a nudge: everything from the checklist is now in place and has been stable for a few days — Glama listing with score badge resolving, Dockerfile + CI proving the image answers MCP `initialize`, and the entry offers `npx -y effectfence` matching neighbouring entries. Since the last comment the server also shipped prebuilt binaries for all four platforms and the npm package passed 750+ downloads in its first week. Happy to adjust anything else the list needs.

PR #11537, opened 2026-08-05
Add ninetails-io/gnucash-mcp to Finance & Fintech 🤖🤖🤖

gomezfox:
Adds gnucash-mcp — full double-entry accounting on local GnuCash books (SQLite), read and write. Modular tool surface via role presets. Python, local-only, MIT. Listed on Glama with a passing build and score badge per the listing requirements. Placed alphabetically by author in Finance & Fintech.


gomezfox:
Since filing: now v1.4.4, and the Glama score is AAA. Entry tool counts refreshed to match; still merges clean.

PR #11633, opened 2026-08-06
Add Perfex CRM (official, self-hosted) to Customer Data Platforms 🤖🤖🤖

themesic:
Adds the official MCP server for **Perfex CRM**, the self-hosted PHP CRM, under Customer Data Platforms - next to the other CRM entries such as Pipedrive.

**Disclosure:** I maintain this product.

- Registered in the official MCP registry as `com.themesic/perfex-crm` via a DNS-verified namespace: https://registry.modelcontextprotocol.io/v0/servers?search=perfex
- Transport is Streamable HTTP at `POST /api/mcp`, served from the user’s own Perfex installation. There is no single public endpoint or public repository for the server itself, so the entry links to the product documentation, following the pattern a few existing entries already use.
- 148 tools, filtered per API token, so a read-only token exposes only read tools.
- Free MIT-licensed companion resources are public: code examples in cURL, PHP, Python and JavaScript, a Postman collection and the OpenAPI 3.0 specification, at https://github.com/themesic/perfex-rest-api-examples

Marked with the official-implementation and cloud-service markers per the legend. One line added, nothing else touched.


themesic:
Thanks for the review, and for tightening the listing requirements.

**GitHub link:** fixed in 728b431. The entry now points at the repository, and the
product page moved into the description.

**Glama badge:** I have to flag a structural mismatch rather than ask for time.

This server is not a standalone process. It is an HTTP endpoint built into a
commercial Perfex CRM module, served from each customer's own installation at
`POST /api/mcp`. There is no repository that would boot on its own.

For the Glama check to pass, the image would have to contain Perfex CRM itself,
which is a commercial product I have no right to redistribute, plus the licensed
module, a database and a seeded API token. So this is not a missing Dockerfile,
it is an architecture the check does not cover.

What I can offer instead as evidence that it works:

- Published in the official MCP registry as `com.themesic/perfex-crm`, which
  required DNS-based domain ownership verification.
- Full endpoint reference, including the MCP method list, at
  https://perfexcrm.themesic.com/apiguide/
- Code examples, a Postman collection and the OpenAPI 3.0 specification in the
  linked repository.

Two questions:

1. Would the official registry entry be acceptable in place of a Glama score for
   servers that ship inside a self-hosted commercial application?
2. You mentioned connectors for hosted endpoints. Mine has no single hosted URL,
   since the host is the customer's own domain. Does the connectors listing
   support a host variable?

Happy to withdraw the entry if self-hosted servers of this shape are out of scope
for the list. Just say the word.

punkpeye:
Hi @themesic, thank you for your contribution. I see this PR has been open for 31 days without further updates from you. The Glama submission requirement is a core part of our listing process, and that hasn't been fulfilled. Since we haven't heard back, I'm going to close this PR. You're welcome to re-open it or submit a new one when the server is listed on Glama.ai — just make sure to add the badge to the entry. Thanks again for your work!

PR #11671, opened 2026-08-07
Add zigbee2mqtt-mcp 🤖🤖🤖

alexpfau:
Adds [zigbee2mqtt-mcp](https://github.com/alexpfau/zigbee2mqtt-mcp) under **Home Automation**.

An MCP server for *administering* a Zigbee2MQTT estate rather than just controlling devices. It talks directly to the [Zigbee2MQTT MQTT bridge API](https://www.zigbee2mqtt.io/guide/usage/mqtt_topics_and_messages.html), so it works with any Zigbee2MQTT 1.17+ install regardless of whether Home Assistant, Node-RED or nothing else is in the picture.

Covers the layer those integrations do not expose: link quality, mesh topology, interview state, OTA availability, `permit_join`, bindings, reporting intervals and device options. The flagship tool is a one-call estate health report — offline devices, weak links, low batteries, pending firmware updates and devices that keep rejoining the mesh.

- **License:** MIT
- **Language:** TypeScript, stdio, two runtime dependencies
- **Install:** `npx zigbee2mqtt-mcp` ([npm](https://www.npmjs.com/package/zigbee2mqtt-mcp), published with provenance)
- **MCP Registry:** `io.github.alexpfau/zigbee2mqtt-mcp`

Destructive tools (device removal, OTA flashing, Touchlink, bridge restart) are gated behind an opt-in write mode and additionally require an explicit `confirm` argument, so they are not reachable by default.

Entry added at the end of the Home Automation list, matching the existing append order in that section.


alexpfau:
Thanks — both requirements are now met.

1. **Glama listing:** the server was already indexed at https://glama.ai/mcp/servers/alexpfau/zigbee2mqtt-mcp
2. **Badge:** added to the entry in the latest push.

Your check also surfaced a genuine bug, so thank you for that: the server previously exited fatally when `Z2M_MQTT_URL` was unset, which meant it could never answer `tools/list` for a scanner — or for any client that enumerates tools before the user has configured anything. It now starts unconfigured, serves introspection normally, and returns the configuration error on the first tool call instead. Released in v0.2.3.

I also added a Dockerfile (multi-stage, `node:24-alpine`, ~193 MB) and verified the built container answers `initialize` and `tools/list` over stdio with no configuration present.

PR #11707, opened 2026-08-07
Add 2ools MCP server (Developer Tools)

ryanluriea:
Adds **2ools** to Developer Tools.

- **Repo:** https://github.com/2ools/mcp-server
- **Endpoint:** `https://2ools.app/mcp` (remote, Streamable HTTP)
- **Registry:** published as the domain-verified namespace `app.2ools/mcp`
- **Tools:** 43, each with explicit `readOnlyHint`/`destructiveHint` annotations

**Try it without an account.** `create_public_preview` is genuinely authless — it turns a standalone HTML document the model just wrote into a real shareable URL that expires after 24 hours. No signup, no key.

Everything else uses OAuth 2.1 authorization code + PKCE with least-privilege scopes. Notably, `projects:build` and `projects:approve` are separate: an agent that can produce candidate versions cannot approve its own work unless a human explicitly delegates review authority.

Source never travels through the model transcript — exports and artifact downloads are short-lived signed URLs bound to the action, account and exact version, so a large site does not consume the context window.

One line added, in the Developer Tools section. Thanks for maintaining this list.


punkpeye:

PR #11719, opened 2026-08-08
Add drpwchen/vault-search to Knowledge & Memory 🤖🤖🤖

drpwchen:
Adds one entry to **Knowledge & Memory**.

**[drpwchen/vault-search](https://github.com/drpwchen/vault-search)** — 🐍 Python, 🏠 local, MIT.

Semantic search, related-notes and RAG chat over an Obsidian vault. The reason it is a bit unusual: the Obsidian plugin and the MCP server share a *single* LanceDB + Ollama index, so a vault is embedded once and both the human UI and the agent tooling read the same store. Nothing is sent off the machine.

Disclosure: I am the author. Entry added at the end of the section, matching how the existing entries in that section are ordered (that section is not currently alphabetical). Happy to move it if you would rather I insert it alphabetically.

🤖 Prepared with [Claude Code](https://claude.com/claude-code) — flagged per CONTRIBUTING.md.


drpwchen:
Thanks for the quick review, and for spelling out the Glama requirement.

Closing this one myself rather than leaving it stale. vault-search is deliberately local-first: it indexes an Obsidian vault on the user's own machine and depends on a locally running Ollama, so passing a containerised introspection check would mean shipping a Dockerfile for a configuration nobody actually runs. That is not a good trade for a listing.

No issue with the policy — it makes sense for keeping the list working. If the project ever grows a sensible containerised mode, I will come back and do it properly.

PR #11720, opened 2026-08-08
Update proximo entry: tool count 365 -> 906

john-broadway:
One-number refresh of the existing john-broadway/proximo entry: the tool surface grew from 365 to 906 (all four Proxmox products — VE, PBS, PMG, PDM — on the one audited control plane; count verifiable cold: the README's Verify-in-60-seconds block asks the server itself). Nothing else in the entry changed.


john-broadway:
both bots are firing on a one-line edit.

the diff changes one number on the existing proximo entry: 365 tools -> 906. no line added, so there's nothing to de-duplicate.

glama: proximo is already listed at https://glama.ai/mcp/servers/john-broadway/proximo, and the score badge is already on the entry, in the same spot the other badged entries use — kaka-milan-22/kops sits two lines below it with the same layout.

nothing to fix on my side. say the word if you want it shaped differently.

PR #11789, opened 2026-08-09
Add Unl under Agreements & Coordination 🤖🤖🤖

neilmortimer-kar:
Adds **Unl** under **Agreements & Coordination** — one entry, one line, existing format and emoji legend (🎖️ official · 📇 TypeScript · ☁️ cloud).

**What it is:** a remote MCP server that serves the decisions a person has already settled, with the reasoning behind them, into an agent's context before it acts — so the agent inherits that judgment rather than asking again. It sits in this category rather than a tools category because the thing being coordinated is authority: the shared environment carries aims, precedent and who decided what, across a human and however many agents are working.

**Why the category fits:** the section is for "multi-party decision workflows across humans, agents, and organizations", which is exactly the unit here — a settled decision, its reasoning, and what an agent is entitled to do with it.

- Endpoint: `https://api.unlimitless.ai/mcp` (streamable HTTP, OAuth with PKCE + Dynamic Client Registration)
- Seven tools, each annotated: reads are inert, and every write requires an in-turn human gesture
- Free at launch; works on any MCP client, not one vendor's

The link is the product site rather than a repository — the implementation is closed-source, which is why there is no repo to point at. Happy to move the entry, shorten the description, or drop the 🎖️ if you'd rather reserve it.


neilmortimer-kar:
Closing this one, with thanks to both bots for the clear instructions.

The url-check is right and I had it wrong: I raised this pointing at our product site on the evidence of existing non-GitHub entries in the list, and CONTRIBUTING.md already says the server name should be linked to its repository. The older entries are grandfathered; the rule is the rule.

There is now a GitHub repository for the server at https://github.com/Unlimitless-Inc/unlimitless-mcp, so rather than rewrite this PR in place I am closing it and opening a fresh one against the new link, in the correct alphabetical position. Apologies for the extra round trip, and thank you for maintaining the list.

punkpeye:
Hey! This didn't make it into this list, but if it has a hosted/remote endpoint we're now also curating https://github.com/punkpeye/awesome-remote-mcp-servers for that. PR welcome over there 🙌

PR #11846, opened 2026-08-10
Add safenode-mcp-gateway to Security

sp3ak:
Adds SafeNode MCP Gateway to the Security section, alphabetically after snyk/studio-mcp.

A proxy that evaluates each tools/call against policy before forwarding it; a denial never
reaches the downstream server. MIT, npx-installable, no telemetry.

https://github.com/sp3ak/safenode-mcp-gateway


punkpeye:

PR #11991, opened 2026-08-12
Add Busymail to Communication

vedhavyas:
Busymail is an email client for the IMAP accounts you already have, and it exposes them over MCP. An assistant can read and search your mail, archive and pin things, and draft replies, with anything it writes queued for you to approve before it sends.

It's a hosted service rather than a repo, so the link goes to the setup docs. It's also on the official MCP registry as `app.busymail/busymail`.

Added to Communication in alphabetical order.


vedhavyas:
Fair enough - Busymail is a hosted service with no public repo, so it can't meet the GitHub-only rule. Closing. Thanks for the pointer to the connectors directory, that's the right home for it.

PR #12002, opened 2026-08-12
Add meet-live-assist: your own agent inside a live Meet/Zoom call 🤖🤖🤖

krystiangw:
Adds one entry to **Communication**, in alphabetical order between `korotovsky` and `lanchuske`.

[krystiangw/meet-live-assist-extension](https://github.com/krystiangw/meet-live-assist-extension) - a Chrome extension plus a local Node server. The extension reads the on-screen captions of a Google Meet or Zoom call, the server exposes them over MCP, and the agent you already run answers in a side panel while the call is still going, posts to the meeting chat, and reads screenshots of a shared screen.

It is local-first rather than another meeting bot: nothing joins the call, transcripts and screenshots are files on the user's own disk, there are no accounts and no service of ours to talk to. It needs an MCP client attached to be useful, which the entry says plainly so nobody installs it expecting a standalone notetaker.

Server: `npx meet-live-assist-server` (Node 20+). Source is public; licence is PolyForm Internal Use 1.0.0, so it is free to read, run and use at work, and not to resell.

Markers: 📇 JavaScript · 🏠 local · 🍎 🪟 🐧


krystiangw:
Thanks. Working on the Glama listing now.

One thing worth flagging for the check: this server is the MCP half of a Chrome extension, so its tools talk to a bridge server on the user's own machine. It answers `initialize` and `tools/list` on its own with no bridge running, no token and an empty environment, which I have verified, so introspection passes. A `Dockerfile` is now in the repo for that purpose - it deliberately does nothing else, because from inside a container the tools cannot reach the host loopback they exist to talk to.

I will add the badge here once the Glama listing is up.

krystiangw:
Badge added - Glama had already indexed the repo, so the listing exists at `krystiangw/meet-live-assist-extension`.

Note for whoever runs the check: Glama currently reports 0 tools for it. The server exposes 13 over stdio and answers `initialize` and `tools/list` with no bridge server, no token and an empty environment - I verified that against the published package. There is now a `Dockerfile` in the repo for exactly this, so a container run should introspect fine.

PR #12018, opened 2026-08-12
Add Blahaj-gif/Finance-mcp (Finance & Fintech) 🤖🤖🤖

Blahaj-gif:
### What it is

A brokerage MCP server for Webull, Saxo and Interactive Brokers behind one
broker protocol, plus parsed SEC filings, options analytics and the BLS/Fed/BEA
macro calendar.

### Why it may be worth a line

**The assistant can draft an order. It cannot send one.** Drafts go to a local
queue; sending requires a broker preview and a human click in a Streamlit
dashboard. That path does not exist on the tool side, so no prompt reaches it —
and **32 of 36 tools declare `readOnlyHint`**, with `cancel_order` the only
`destructiveHint`, so a client can enforce that rather than take the README's
word for it.

Filings are parsed rather than forwarded: Form 4 with transaction codes and
10b5-1 status, 8-K by item code, 13F, 13D/G, NPORT. **13F holdings are
reconciled against the entry count and total value the filing declares on its
own cover page** — checked across ten institutions and 82,701 rows, matching to
the dollar, and the result is reported in the tool output so a reader is told
when it does not add up.

### Stated plainly in the entry, because it matters

Only the **Webull** adapter has been run against a live account. The Saxo and
IBKR adapters are written from published REST documentation, carry
`verified = False`, and announce it in every result they produce. The project is
asking for someone with a simulation or paper account to close that gap.

### Housekeeping

- MIT, Python 3.10 and 3.12 on CI, **1373 tests**
- CI also builds the wheel and the Docker image and runs a real MCP handshake
  against each, offline — the source tree always has every module, which is why
  a missing subpackage in a published wheel is invisible to a test suite
- Glama: **A / A / A**, Tool Definition Quality 4.1/5 — badge in the entry
- Published as `hitl-finance-mcp` on PyPI and listed on the official MCP
  registry as `io.github.Blahaj-gif/hitl-finance-mcp`
- Added to the top of Finance & Fintech, matching the section's existing order

Opened by an agent on the maintainer's behalf, hence the 🤖🤖🤖 in the title per
CONTRIBUTING.

<sub>Updated since opening: three tools were removed upstream (two superseded by
the EDGAR and XBRL parsers, one whose job is a parameter on another), so the
count went 39 → 36. The entry and this description now match the server.</sub>


Blahaj-gif:
Badge added — thanks for the pointer.

Worth noting for anyone hitting the same thing: the server was **already listed
on Glama**, auto-discovered from GitHub rather than submitted, so a manual
submission came back as "MCP server already exists for this repository". The
listing and badge were live the whole time:
https://glama.ai/mcp/servers/Blahaj-gif/Finance-mcp

It builds and starts under introspection with no credentials and no network —
there is a `Dockerfile` in the repo and CI runs the image with `--network none`
and a real three-message handshake on every push, because a server that needs
the internet to list its own tools fails in exactly the sandbox a directory
runs it in.

PR #12030, opened 2026-08-12
Add django-orm-lens under Developer Tools

FROWNINGdev:
Adds `FROWNINGdev/django-orm-lens` to **Developer Tools**.

Static-analysis MCP server for Django models. It parses `models.py` with Python's `ast` module, so it needs no database connection, no Django boot and no credentials — all tools are read-only. Useful when an agent has to reason about a schema it cannot query.

Tools exposed: schema and relations, migration/schema drift, cross-function N+1 detection, and the blast radius of a change to a model.

- Repo: https://github.com/FROWNINGdev/django-orm-lens (MIT, 10 contributors)
- PyPI: https://pypi.org/project/django-orm-lens/
- Already published in the official MCP registry as `io.github.FROWNINGdev/django-orm-lens`

Flags used: 🐍 Python, 🏠 local service, 🍎 🪟 🐧 cross-platform.

I left the Glama badge off deliberately rather than hand-crafting a badge URL for a server that may not be indexed yet — happy to add it if you would like it in the entry.


FROWNINGdev:
The duplicate check is right — `FROWNINGdev/django-orm-lens` is already listed, and the existing entry is more complete than the one I proposed (it enumerates all nine tools and carries the Glama badge). I missed it because I searched the section headings rather than the full 1.3 MB README. Closing; nothing here to merge. Sorry for the noise.

PR #12105, opened 2026-08-13
Add AgentRamp to Aggregators 🤖🤖🤖

Rexcraig:
Adds AgentRamp to the Aggregators section.

AgentRamp generates a hosted MCP server from a product's public API documentation, so a
SaaS can be called as tools by an agent without writing MCP code. This is the same shape
as `wegotdocs/open-mcp`, already in this section, with the difference that AgentRamp hosts
and monitors the resulting endpoint rather than emitting a local server.

Two things I want to flag rather than have a maintainer find:

- It is a hosted service, not an open-source repository, so the link goes to the docs. I
  see a small number of existing entries doing the same (BrainGrid, shadcn/studio) but
  tell me if you would rather not, and I will close this.
- There is a free tier (1 product, no card) so the service is testable without paying.

Endpoint format and a worked example are in the docs. Happy to adjust the wording,
badges, or section placement.


Rexcraig:
Closing this for now. The url-check is right: AgentRamp is a service that generates MCP servers rather than a server hosted on GitHub, so a docs link was never going to fit the listing rules.

Rather than argue the edge case, we'll come back with an actual MCP server in its own repo that meets the requirements, including the Glama listing. Thanks for the clear automated feedback, it saved a round trip.

PR #12207, opened 2026-08-15
fix(alphai-mcp): correct free tier — 100 calls/day, not per hour 🤖🤖🤖

makeev:
Corrects the free-tier description on our own entry (`makeev/alphai-mcp`).

The line currently reads `free 100 calls/hour`. The actual free tier is **20 requests/minute and 100 calls/day** — a per-day volume cap, not per-hour. As written the listing overstates the free allowance by 24x, so anyone taking it at face value runs into a rate limit they were told they would not hit.

One line changed, no other edits.


makeev:
For whoever picks this up in manual review: this is an in-place edit of the existing `makeev/alphai-mcp` entry, not a new listing — hence the bot's `duplicate` flag. The Glama score badge is already on the line (unchanged); the only change is the free-tier text, `100 calls/hour` → `20 req/min and 100 calls/day`, which is what the server actually enforces. Still one line, still mergeable-clean.

PR #12230, opened 2026-08-15
update GBLIN-MCP entry: lead with what it does

gblinproject:
The old wording described the project in our own words ("treasury standard") rather than what an agent is looking for. Same entry, now stating the actual capability: free market risk regime reads, signed risk attestations, and the collateral-backed index for idle USDC, plus the install command like neighbouring entries.


gblinproject:
Both bots are looking at the same line from different angles, so one comment for both.

The duplicate check is firing on our own entry. This PR doesn't add a server: it rewrites the description of the gblinproject/GBLIN-MCP line that's already in the list. One line changed, nothing added — the diff shows it.

The Glama badge is already on that line and this PR keeps it untouched. The server is listed and responding at glama.ai/mcp/servers/gblinproject/GBLIN-MCP.

Why rewrite it at all: the old text opened with the treasury product, which is the part almost nobody wants first, and buried the part that's free — market risk regime, calm/elevated/crash, no API key and no wallet. The new line leads with that.

Also pushed a correction to the branch just now: the tool count says 13, not 10. Checked by loading the published package rather than counting by hand.

PR #12237, opened 2026-08-16
Fix broken Star History chart in README

OctoBored:
The Star History chart at the bottom of the READMEs is currently broken and no longer renders, which prevents visitors from seeing the project's growth. This PR updates the chart links across README.md and all localized README versions (fa-ir, ja, ko, pt_BR, th, zh, zh_TW) so the star history displays correctly again.

A proper star history is valuable for this project as it showcases the community growth at a glance, so fixing this is important.


OctoBored:
Not applicable

PR #12286, opened 2026-08-16
Add otto (paper-only, agent-operated financial terminal) 🤖🤖🤖

0Smallcat0:
Adds otto under Finance & Fintech.

- Local-first financial terminal operated by an AI agent over MCP (stdio, stdlib-only server, auto-starts its own backend; state stays on the user's machine).
- Paper trading only: live trading, credential entry, and code execution are structurally unreachable through the tool surface (`pytest tests/test_m16_live_safety.py` — every live-action endpoint is rejected with no side effects, and the disabled-endpoint registry covers all actions).
- Every trading decision goes through a pre-registered, programmatically graded judgment ledger: stance, invalidation level, and horizon are written before the scored window opens and cannot be edited afterward.
- Ships a reproducible agent-operability benchmark instead of an unmeasured "AI-powered" claim, graded programmatically with no LLM judge — current results in the repo (`evals/EVAL.md`).
- Install (documented today, ahead of the PyPI release): `claude mcp add otto -- uvx --from git+https://github.com/0Smallcat0/otto otto-terminal`

Repo: https://github.com/0Smallcat0/otto


punkpeye:
Thanks for your PR! The entry looks good, but there's one issue: the server needs to be **claimed** on Glama. You can do this at https://glama.ai/mcp/servers. The quality score has already been evaluated, so once you claim it, we'll be good to merge. Let us know if you need any help!

PR #12305, opened 2026-08-17
Add Cost of Work Index (hirenika) MCP server 🤖🤖🤖

kindrat86:
# Add Cost of Work Index (hirenika) MCP server

## What

One new entry in Finance & Fintech for the [Cost of Work Index](https://github.com/kindrat86/hirenika/tree/main/packages/cost-of-work-mcp) MCP server.

It answers the question "how much does one unit of a back-office task cost when a person does it versus when software does it" for 13 tasks across the US, Greece and Ukraine, with a cited official statistic behind every hourly wage (BLS OES, Eurostat labour cost, Ukrstat). Human time per task is labelled as an estimate with its basis stated per task. Rows where the human is cheaper are kept rather than filtered, and the caveats ship inside the data payload.

## Install

```
npx -y nika-cost-of-work-mcp
```

Node 18+, no API key, dataset bundled so it answers offline.

## Checklist

- [x] Entry inserted alphabetically in the Finance & Fintech section (before `kindrat86/mcp-deal-flow-signal`)
- [x] Description follows the section's one-liner style with install command
- [x] No Glama badge included (no listing there yet)
- [x] License CC BY 4.0, data snapshot generated from the live index (no drift)


kindrat86:
Thanks for the pointer. Updates:

- Entry now links to the standalone server repo `kindrat86/cost-of-work-index` (source, Dockerfile and glama.json committed there) and carries the Glama badge in the requested format.
- The Glama listing itself is pending the manual submit at glama.ai/mcp/servers; the `npx` install resolves to the npm package `nika-cost-of-work-mcp`, which is in flight.

Local verification done: initialize handshake + tools/list + a real tool call all pass against the built server (5 tools, dataset bundled, no API key).

PR #12308, opened 2026-08-17
Add sonlenef/storepilot-mcp to Developer Tools

sonlenef:
Adds [StorePilot](https://github.com/sonlenef/storepilot-mcp), an MCP server for managing an app portfolio across Google Play and App Store Connect from one tool set.

What makes it different from the other store servers already listed: it reads Play's private Cloud Storage reports bucket, which is the only place installs, ratings and earnings exist — there is no REST API for them — and it treats both stores as one portfolio rather than one app at a time.

- Python, runs locally over stdio, `pip install storepilot`
- MIT, CI on 3.11–3.13, 503 tests
- Listed in the official registry as `io.github.sonlenef/storepilot-mcp`

Writes are gated: each one previews first and requires an HMAC-keyed, single-use confirmation token, and production releases are forced into a staged rollout.

The README is explicit about what has not been verified against real store data yet, rather than implying everything is proven.


punkpeye:
Thanks for the submission! A couple of things are needed before we can merge:

1. **Claim your server on Glama**: Go to https://glama.ai/mcp/servers, find your server, and claim it (be sure to verify ownership).
2. **Add the Glama badge** to your entry: once claimed and scored, add the badge in the format `[MCP Score](https://glama.ai/mcp/servers/…)`. The badge should be placed after the GitHub URL and before the emojis.

Once those are done, let me know and I'll take another look!

PR #12356, opened 2026-08-17
cachly-mcp: correct tool count (84 → 122)

HeinrichNebula:
The cachly-mcp entry has listed 84 tools since it was added. The server ships **122** today — the count in `sdk/mcp/src/tools.ts` and the project README agree on that number, so the listing understated it by roughly a third.

Also adds `EU-hosted`, which is the one property that sets it apart from the other memory servers in the same section (the neighbouring entries are all local-first).

No other changes.


HeinrichNebula:
Thanks for the automated review — I think both flags are false positives here, and I want to save a maintainer the click.

**On `duplicate`:** this PR does not add a listing. It edits the existing `cachly-dev/cachly-mcp` line in place — the diff is +1/−1 on the same line, one word changed:

```diff
-84 tools — Team Brain, semantic BM25+ search, …
+122 tools — Team Brain, semantic BM25+ search, … EU-hosted.
```

The bot appears to treat any PR touching a server line as a new submission, which is why it found the entry "already listed" — it is, and that is exactly the line being corrected.

**On `missing-glama`:** the Glama score badge is already present in that line and untouched by this change:

```
[![cachly-dev/cachly-mcp MCP server](https://glama.ai/mcp/servers/cachly-dev/cachly-mcp/badges/score.svg)](https://glama.ai/mcp/servers/cachly-dev/cachly-mcp)
```

**Why the correction:** the entry has said 84 since it was added. The server ships 122 tools today — `sdk/mcp/src/tools.ts` and the project README agree on that number, so the listing understates it by about a third. Happy to drop the `EU-hosted` addition if you would rather keep the diff to the number alone.

PR #12394, opened 2026-08-18
Add Flicker MCP server (Finance & Fintech)

grep-o:
Flicker is an official, hosted (Rust/rmcp) MCP server exposing crypto/stock/FX market data, Flicker analysis zones, and the user's own multi-exchange portfolio, watchlist and alerts — read-only apart from watchlist/alerts/notification preferences; it does not place trades.
Endpoint: https://api.flicker.finance/mcp (Streamable HTTP, OAuth required on every request, so unauthenticated hits return 401).


punkpeye:

PR #12434, opened 2026-08-18
Add Vidmoat — timeline video editor with visual feedback to the agent 🤖🤖🤖

fredabila:
Adds **Vidmoat** to Multimedia Process.

Vidmoat is a hosted timeline video editor exposed over MCP. What makes it worth a line in this list: **every edit returns rendered preview frames as images**, alongside a machine-readable `lint[]` (overlapping text, offscreen elements, unreadable font sizes). The model can look at the frame it just produced and correct itself, rather than writing timeline JSON blind and hoping.

Beyond that it does the ordinary editor work an agent needs end to end — import footage or search stock, cut and arrange clips, captions from real word timings, effects and keyframes, then render an MP4.

- Remote Streamable HTTP: `https://api.vidmoat.com/api/mcp`
- Registered in the official MCP registry as `com.vidmoat/vidmoat`
- Requires a Vidmoat account; MCP access is on the Studio plan (stated in the entry so nobody finds out after connecting)

One line added, in the existing `v-` cluster, following the surrounding format and the legend (🎖️ official, 📇 TypeScript, ☁️ cloud, 🍎🪟🐧).

Title carries the `🤖🤖🤖` marker per CONTRIBUTING.md — this PR was prepared by an agent on behalf of the maintainer.


fredabila:
Thanks for the quick automated review — both points are fair and I'm withdrawing rather than working around them.

Vidmoat's MCP server is a hosted remote endpoint (`https://api.vidmoat.com/api/mcp`) with a closed source repo, so there is no GitHub URL I can point at, and no Dockerfile for Glama to boot and introspect. I do have a public repo under this account, but it's an unrelated fork and listing it here would misrepresent what it is — so that's not an option either.

Following your P.S. instead and pursuing the hosted-endpoint route at https://glama.ai/mcp/connectors, which is clearly the right home for this. Happy to come back with a proper PR if we ever open-source the server itself.

Sorry for the noise, and thanks for keeping the bar where it is — the list is more useful for it.

PR #12466, opened 2026-08-19
Add Gliana-Labs/gliana-mcp — pay-per-call AI over HTTP 402 🤖🤖🤖

Zetakai:
Adds GlianaAI to **Aggregators**, alphabetically between `glenngillen/mcpmcp-server` and `gpu-bridge/mcp-server`.

One MCP server for 90+ AI models (LLM chat, image, video, music, speech) plus utility and data tools — scraping, screenshots, OCR, face matching, crypto and FX rates. There is no signup and no API key: HTTP 402 is the only gate, and each call settles from the caller's own wallet in USDC on Base, Solana, BNB Chain or Algorand. The tool list is read live from the gateway rather than hardcoded, so it does not go stale as models are added or retired.

- Server: https://github.com/Gliana-Labs/gliana-mcp
- npm: `npx -y gliana-ai-mcp` (also hosted at `https://mcp.glianalabs.com`)
- Docs: https://ai.glianalabs.com/docs

Single line added, existing format and emoji legend followed (📇 TypeScript, ☁️ cloud, 🏠 local, plus platforms).


Zetakai:
Thanks — both done.

**Badge added** after the description, using the Glama path `Gliana-Labs/gliana-mcp`. The server is already indexed at https://glama.ai/mcp/servers/Gliana-Labs/gliana-mcp.

**Dockerfile added** so the introspection checks can build and start it: https://github.com/Gliana-Labs/gliana-mcp/blob/master/Dockerfile

Worth noting for the check — the handshake needs **no credentials**. Discovery tools read a public catalog; only a paid call needs a wallet key, and that stays on the user's machine. So the image ships with no secrets, runs as non-root, and exposes no port (stdio transport).

Verified locally against the built image:

```
initialize  -> gliana-ai 0.5.2
tools/list  -> 7 tools (list_models, get_price, get_schema, generate, list_tools, tool, recipe)
```

The probe also turned up a bug worth having: `serverInfo.version` was hardcoded at 0.5.0 while the package shipped 0.5.2, so every client was told the wrong version. It now reads package.json at startup.

Also noted the connectors suggestion — there is a hosted endpoint at `https://mcp.glianalabs.com`, so I will list it under https://glama.ai/mcp/connectors as well.

Zetakai:
The Glama evaluation has come through — the server now carries a quality score:

**License A · Quality A · Maintenance B** — https://glama.ai/mcp/servers/Gliana-Labs/gliana-mcp

That was the outstanding item from the badge check above.

PR #12468, opened 2026-08-19
Add VidWords YouTube MCP server 🤖🤖🤖

haljishi:
Adds **[haljishi/vidwords-mcp](https://github.com/haljishi/vidwords-mcp)** to **Multimedia Process**, next to the other "let an agent actually read a video" servers (klaket, mcp-video-analyzer, clipy).

A hosted Streamable HTTP server (`https://vidwords.com/mcp`) with nine tools. What makes it different from the several YouTube-transcript servers already on the list:

- **`search_transcript` is the flagship, not `get_transcript`.** It returns only the stretches that answer the question, each with a `youtube.com/watch?v=…&t=…s` deep link, instead of dumping 20,000 words into the context window.
- **It reads the frames, not only the captions** — slides, charts, demos and on-screen text that is never spoken aloud.
- **Citations are verified before you see them.** A visual claim must match a frame that was actually recorded and a spoken claim a real transcript segment; anything that fails is dropped, and when nothing survives the answer says the evidence is insufficient rather than guessing.

Official implementation (🎖️), remote/cloud (☁️), works on any OS. Auth is a token header or OAuth for claude.ai and ChatGPT. Free tier, no card.

Also published in the official MCP registry as `com.vidwords/youtube`.

Checked for duplicates first — no existing entry mentioned vidwords. Placed alphabetically before `huseyinstif/klaket`.


haljishi:
Thanks — both steps addressed, with one wrinkle worth flagging since it applies to any hosted server with authenticated discovery.

**Step 2 (badge):** added after the description, pointing at `haljishi/vidwords-mcp`.

**Step 1 (Glama check):** this is a remote-only server (`https://vidwords.com/mcp`), and pointing a checker straight at it would not have worked — the endpoint returns `401` on `initialize` itself, so an unauthenticated introspection never reaches `tools/list`.

The obvious fix would have been to allow unauthenticated `initialize`, and I deliberately did not do that: the `401` is what carries `WWW-Authenticate` with the resource-metadata URL, and it is what starts the OAuth flow for claude.ai and ChatGPT. Loosening it to satisfy a crawler would put the connector those clients rely on at risk.

So the repository now ships a **stdio proxy plus a Dockerfile** instead, in the shape `SubDownload/subdownload-mcp` uses. The nine tool schemas are declared inline, so `initialize` and `tools/list` answer with **no credentials at all** and the upstream is never contacted during introspection; only `tools/call` forwards, with the token from `VIDWORDS_API_TOKEN`, and a call without one returns a readable error rather than breaking the handshake.

Verified against the built image before pushing:

```
$ docker run --rm -i vidwords-mcp   # no environment whatsoever
initialize  -> {"name":"vidwords-youtube","version":"1.0.1"}
tools/list  -> 9 tools
```

`glama.json` is in the repo root. The badge will render blank until the Glama listing is created — that submission needs an account action on our side and is in progress. Happy to adjust the placement or the entry if anything here does not match what you expect.

haljishi:
Following up on the score check — Glama has now evaluated the server, so the badge in the entry renders a score rather than a blank.

Breakdown at https://glama.ai/mcp/servers/haljishi/vidwords-mcp/score:

- **Server Coherence — A** (disambiguation 5/5, tool count 5/5, completeness 5/5)
- **Tool Definition Quality — A**, 4.4/5 average across 9 of 9 tools scored
- **Maintenance — B**
- **License — A** (MIT), valid `glama.json`, author verified, profile 83% complete

The Dockerfile route worked as intended: `initialize` and `tools/list` answer with **no credentials at all**, because the stdio proxy declares the nine schemas inline and never contacts the hosted endpoint during introspection. That left the `401` + `WWW-Authenticate: Bearer resource_metadata=…` on `https://vidwords.com/mcp` intact, which is what starts the OAuth flow for claude.ai and ChatGPT — so satisfying the checker did not cost the connector those clients rely on.

Both listed requirements should be satisfied now. Happy to adjust the entry or the description if anything else is needed.

PR #12483, opened 2026-08-19
Update README.md

Tanmay-rightaichoice:


Tanmay-rightaichoice:
Happy to add the badge, but I don't think one can exist for this entry — flagging that rather than quietly ignoring the check.

This is a remote streamable-HTTP server, not a stdio server in a repo, so there's no Dockerfile to submit to Glama. It is listed there, as a Connector: https://glama.ai/mcp/connectors/com.rightaichoice/right-aichoice — and connectors don't seem to get score badges. I tried both badge URL formats and they 404.

It's also published in the official MCP registry as com.rightaichoice/mcp with DNS verification, if that's a useful signal in place of the Glama score.

Are remote connectors eligible under the new requirement, or would you rather I close this?

punkpeye:
Thanks for the explanation. Connectors listed on Glama as remote HTTP servers do not receive a score badge in the traditional sense — you're correct about that.

Our contribution rules require that listings include a Glama badge, which is generated when a server is submitted as a standard MCP server (with a Dockerfile) at https://glama.ai/mcp/servers. Since your server is a remote streamable-HTTP server and appears only as a Connector, it doesn't produce a badge URL.

To proceed, you have two options:

1. **Submit the server as a standard MCP server** on Glama by adding a Dockerfile (even if it just starts and responds to introspection). Once Glama evaluates it, a badge will be generated and you can add it to the entry.
2. If that's not feasible, I'd suggest submitting your server to https://glama.ai/mcp/connectors directly — those listings are separate from the main curated list here and don't require a badge.

Alternatively, if you believe the rule should accommodate connector-based servers, we can refer this to a human maintainer for review. Let me know how you'd like to proceed.

PR #12523, opened 2026-08-20
Add license-guard to Legal 🤖🤖🤖

rccaoki-wq:
Adds one server under **Legal**, in alphabetical order (between `ptrinh` and `smilemin07`). One line, one file.

### What it does

Determines whether an open source dependency creates a legal obligation, given how the project ships. The same license yields different answers for hosted SaaS, distributed binaries, on-premises delivery, internal-only use, and published libraries — and build-time dependencies carry no distribution obligation at all.

AGPL-3.0 is the clearest case: its section 13 obligation attaches when users interact with the software over a network, so hosting triggers it while purely internal use does not. GPL works the opposite way, attaching on distribution. A scanner that reports "AGPL detected" without knowing which of these you are doing is telling you almost nothing.

It reads lockfiles so transitive dependencies are covered — which is where problematic licenses usually arrive, rather than in dependencies you added directly.

### Why Legal

It sits next to `ark-forge/mcp-eu-ai-act`, a compliance scanner for codebases. This is the same shape, for software license obligations.

### Verified before opening this PR

- Repository link returns 200
- MCP endpoint responds to `tools/list` (Streamable HTTP, stateless, no auth)
- Listed in the official MCP registry as `io.github.rccaoki-wq/license-guard`
- All six lockfile formats named in the entry were exercised against the live endpoint

Apache-2.0. Not affiliated with any registry it reads from.

### Disclosure

Opened by an automated agent, using the `🤖🤖🤖` opt-in described in CONTRIBUTING.md. A human owns the project and approved this submission.


rccaoki-wq:
Both steps are done.

**1. Listed on Glama** — it was picked up automatically from the official MCP registry and from GitHub, so no manual submission was needed.

- Server: https://glama.ai/mcp/servers/rccaoki-wq/license-guard — `A license`, `B maintenance`
- Connector: https://glama.ai/mcp/connectors/io.github.rccaoki-wq/license-guard — **Status: Healthy**, last tested 2026-08-20 07:30

This is a remote server with no local package, so there is no Dockerfile to add. Glama's connector health check exercises the live endpoint directly, which I believe covers the "starts and responds to introspection requests" requirement. Glama also independently filed it under *Legal & Compliance*, which matches the section this PR targets.

**2. Badge added** — placed right after the repository link and before the emoji, matching the ~1,995 existing entries rather than the literal wording in the automated message. Say the word if you would rather have it after the description.

Still a single line, single file. All three URLs in the entry verified 200 before pushing.

rccaoki-wq:
Understood — the quality score needs Glama to run the server, and until now it couldn't: this was a remote-only Cloudflare Worker, so Glama showed *"This server cannot be installed"* and left quality ungraded.

I've added a Dockerfile and a stdio entry point so it can be built and evaluated:

- `Dockerfile` — bundles the server into a single file with esbuild, runs as non-root, no runtime toolchain needed
- `src/local/stdio.ts` — stdio transport, same engine and same three tools as the hosted version

Verified with the official MCP SDK client against the real process (`npm run e2e:stdio`): connect, `tools/list`, and tool calls all behave identically to the hosted endpoint. Also checked that stdout carries only MCP messages, with logging on stderr.

The remaining steps (claim the server, configure the build spec, Deploy, Make Release) happen in Glama's admin UI, so the repository owner is doing those. I'll update this PR once the score lands.

Worth noting the local build turned out to be useful on its own: a compliance tool that requires sending your lockfile to a third-party service is a hard sell for some organisations. Run locally, only package names and versions go to the public registries — the manifest never leaves the machine.

rccaoki-wq:
The Glama quality score has landed: **A**.

- Server: https://glama.ai/mcp/servers/rccaoki-wq/license-guard — `A quality`, `A license`, `B maintenance`
- The score badge is in the entry and now returns a graded SVG.

What changed since the last comment: the Dockerfile and stdio entry point are built and released on Glama, so the server can actually be launched and its tool definitions evaluated — that is what the quality grade required.

I also updated the entry to reflect it:

- Added 🏠 alongside ☁️ — it now runs locally over stdio, not only as a hosted service.
- Noted that the local path keeps the manifest on the user's machine (only package names and versions go to public registries).

Ready for review.

PR #12640, opened 2026-08-22
Add BoardRepo (hardware designs over remote MCP) to Embedded System 🤖🤖🤖

rrenamed:
Adds **BoardRepo** to Embedded System.

BoardRepo is the home for your PCB projects, plus every public design, readable by your AI. Connect it to an assistant and it reads hardware designs themselves rather than guessing from memory: the schematic and the nets that connect it, the bill of materials, the files in the project, and the public library of real-world boards other people have published. KiCad and Altium today, and the list of formats is growing.

- Endpoint: `https://boardrepo.com/mcp` (Streamable HTTP)
- Setup: https://boardrepo.com/connect
- Official MCP registry: `com.boardrepo/boardrepo`
- Auth: OAuth 2.1 with dynamic client registration, so there is no API key or client secret to paste
- Tools: `search_boards`, `get_board`, `read_schematic`, `get_bom`, `list_board_files`, `read_file`. All read-only, all annotated with `readOnlyHint` and `openWorldHint`.

Placed in Embedded System alongside the other hardware-facing servers such as `octoco-ltd/sheetsdata-mcp`. Listed by name rather than `owner/repo` because the server is hosted and closed-source, matching the existing convention for entries like BrainGrid MCP and shadcn/studio MCP.

One line added, no other changes.


rrenamed:
Withdrawing this for now, no action needed from maintainers.

Two reasons: the server's repository is private, so the GitHub-only URL requirement can't be met as things stand, and our tool surface is actively changing this week, so the description would be out of date on arrival. We'll resubmit once both are settled. Thanks for the bot feedback, it caught the URL issue immediately.

PR #12642, opened 2026-08-22
🤖🤖🤖 Add arthiqlabs/vedaksha to Spirituality & Esoterica

arthiqlabs:
Adds [arthiqlabs/vedaksha](https://github.com/arthiqlabs/vedaksha), a self-hosted Rust MCP server for Vedic astrology computation.

- **17 deterministic tools** — natal charts, dashas, vargas, shadbala, ashtakavarga, panchanga, drishti, bhavas, transits, gochara, muhurta search, synastry, composite, and property-graph export.
- **Clean-room** — every algorithm carries a source citation to a primary paper or treatise; the lunar theory and the sidereal surface each ship an audit directory that re-derives their values.
- **Local, no API key** — `cargo install vedaksha-mcp`, stdio or self-hosted HTTP. The analytical ephemeris compiles to constants, so it needs no data files.
- Positions validated against JPL Horizons: 0.103″ mean over 15,350 comparisons, 1900–2025.

One entry, placed alphabetically between `adityarya24/astro-skill` and `astroway/astroway-mcp` in Spirituality & Esoterica. Emoji: 🦀 Rust, 🏠 local, and 🍎 🪟 🐧 — macOS and Linux binaries are built in CI, and the Windows target compiles (`cargo check --target x86_64-pc-windows-msvc`).

Disclosure: I maintain this project.


arthiqlabs:
Thanks — good catch. The server isn't graded on Glama yet, so I've removed the badge rather than ship one that points at an ungraded entry. The line is now just the repo link, emoji and description, matching the many entries here that carry no badge.

Happy to add it back in a follow-up once it has a real score.

arthiqlabs:
Both requirements are now met, and I've re-added the badge.

The server is listed on Glama at https://glama.ai/mcp/servers/arthiqlabs/vedaksha with a Dockerfile added there directly. The build test passes and the server starts and responds to introspection — all 17 tools are enumerated on the listing, and it now shows Install Server / Try in Browser rather than "cannot be installed".

Badge restored after the repo link, matching the placement used by the neighbouring entries in this section.

For transparency on the earlier exchange: I removed the badge a few hours ago because the server was genuinely ungraded at that point and the badge pointed at nothing meaningful. Creating the Glama release fixed that, so it's back.

One caveat you may want to know about, since it affects the badge's appearance rather than its validity: the licence facet reads F for us and for every Business Source License project, because Glama takes it from GitHub's licence detection and GitHub has no BUSL entry — hashicorp/terraform, vault, cockroachdb/cockroach and getsentry/sentry all report `other` for the same reason. I've reported that to Glama separately. Quality is still grading; it runs asynchronously after a release.

PR #12736, opened 2026-08-23
Add markdown-to-whatsapp (Communication)

drsound:
Adds an MCP server that converts Markdown into WhatsApp's formatting syntax, under **Communication**.

The use case is agents that write to WhatsApp: they emit Markdown, WhatsApp doesn't read it, and the message arrives full of literal asterisks. This is a formatting converter — it does not send messages, so it does not overlap with the WhatsApp sender servers already listed.

The non-trivial part is tables. A WhatsApp bubble fits a fixed number of monospace characters, so the server draws each table to that width: it removes padding, then borders, then word-wraps cells, and falls back to a bulleted list only when no box fits. East Asian wide characters and emoji are measured at their real display width so alignment survives them.

stdio, no API key, no network access. `npx markdown-to-whatsapp mcp`. MIT, tests in CI.

Entry inserted in alphabetical order within the category.


drsound:
The Glama evaluation has completed since the bot last checked. The server now has a quality score of **4.6, tier A** (Server Coherence 4.8, Tool Definition Quality 4.5), and the licence detection resolved to MIT, so the earlier F row is gone.

Score page: https://glama.ai/mcp/servers/drsound/markdown-to-whatsapp/score

The badge is already in the PR. Anything else needed from my side?

PR #12878, opened 2026-08-25
Add SolveGate MCP server 🤖🤖🤖

ImErdis:
Adds one server to **Browser Automation**, in alphabetical position between `softvoyagers/pageshot-api` and `swimmwatch/cloakbrowser-mcp`.

### What it is

An MCP server for dealing with Cloudflare Turnstile — the checkbox-or-nothing widget that sits in front of a lot of login and signup forms. It gives an agent three tools:

- **`inspect_page`** — point it at a URL and it tells you whether that page actually uses Turnstile, what its sitekey is, and how the widget is set up. **This one needs no API key and no account.**
- **`solve_turnstile`** — clears the challenge and hands back a token.
- **`get_solve`** — fetches a result you asked for earlier.

### Why the free tool is listed first

The most common mistake an agent makes here is reaching for a solver when it should have looked first. Three situations account for most of it: the page turns out to carry reCAPTCHA or hCaptcha, which this does not solve; or it is a full-page Cloudflare interstitial rather than an embedded widget, which is a different thing entirely; or the widget is built by scripts after the page loads, so there is nothing to read. `inspect_page` answers all three in about a second, for nothing, and quite often that is the whole answer.

It is also deliberately honest about one thing it *cannot* tell you: which mode a widget is in. That setting lives in the site owner's Cloudflare account, not in the page, so no tool can read it off the source — and the server says so rather than guessing.

### Details

| | |
|---|---|
| Package | [`solvegate-mcp`](https://www.npmjs.com/package/solvegate-mcp) on npm, `npx -y solvegate-mcp` |
| Source | https://github.com/SolveGate/solvegate-mcp |
| Licence | MIT |
| Language | TypeScript |
| Legend | 🎖️ official · 📇 TypeScript · ☁️ cloud service |

Checked before opening this: the entry sits in correct alphabetical order, it is one line, both links resolve, and the published package was installed fresh into an empty directory and driven over stdio to confirm the handshake and all three tools respond.


ImErdis:
Done — the badge is in, placed after the repository link to match the surrounding entries.

A note on the Glama step, in case it saves someone else the confusion: the manual submission came back rejected as *"MCP server already exists for this repository"*. That turned out to be good news rather than bad — Glama's crawler had already indexed the repo, and at the exact commit that added the Dockerfile, so it had everything it needed before the form was ever filled in.

The listing was unclaimed, though, which is what actually needed fixing. For a repository owned by an organisation, Glama proves the claim with a `glama.json` file in the repo root rather than by organisation membership, so that is now committed.

The listing: https://glama.ai/mcp/servers/SolveGate/solvegate-mcp

On the Dockerfile — it starts the server and answers introspection with a **completely empty environment**, which I checked by building the image and speaking to it with no variables set at all. That works because the server's `inspect_page` tool needs no API key, so there is nothing to configure before it can tell you what it does. Happy to adjust anything about the entry if the format is off.

PR #7622, opened 2026-06-08
Add Lyrenth (lyrenth-mcp) to Browser Automation

lyrenth:
Adds lyrenth-mcp under Browser Automation.

Read any URL as a clean AIDocument (Markdown + structure) through
Lyrenth's cached index.

- npm: lyrenth-mcp (https://www.npmjs.com/package/lyrenth-mcp)
- Official MCP registry: com.lyrenth/lyrenth-mcp
- Repo: https://github.com/lyrenth/lyrenth-mcp

Placed alphabetically in Browser Automation, follows the entry format
with the TypeScript and local-service icons.

PR #12956, opened 2026-08-26
Move Lyrenth from Browser Automation to Search & Data Extraction

lyrenth:
I operate this server. The entry was added in June (#7622) under
Browser Automation, which is not what it does: the server automates no
browser. It exposes read tools that fetch any public URL as a clean
AIDocument (Markdown plus structure) through Lyrenth's index, so
Search & Data Extraction is the accurate category.

The entry text is byte-identical; only its location moves. One line
removed, the same line inserted under the Search & Data Extraction
header.


lyrenth:
Both bot findings resolve the same way: this PR moves the existing
entry, it does not add one.

The diff is one line removed (from Browser Automation) and the same
line, byte-identical, inserted under Search & Data Extraction. So the
Glama score badge the first comment asks for is already in the moved
line, and the duplicate the second comment flags is the entry's old
location, which this PR removes. After merge the entry appears exactly
once, in the correct category.

PR #12976, opened 2026-08-26
Add Airlock to Security

cyberbobas:
Adds **Airlock** to the Security section.

Airlock is a runtime security firewall for MCP: it wraps your other MCP servers as a stdio proxy and gates every tool call against a least-privilege policy, blocking secret-path reads, known-collector egress, and download-and-execute. It fits alongside the existing runtime-security-proxy entries in this section (e.g. mastyf.ai).

- Repo: https://github.com/cyberbobas/airlock
- PyPI: https://pypi.org/project/airlock-agent/ (`pipx install airlock-agent`)
- Apache-2.0, Python, Linux/macOS/Windows

One line added to the Security section; format and emoji legend followed.


cyberbobas:
Withdrawing — Airlock isn't a listable MCP server, it's a proxy/firewall that wraps other servers, so it can't pass the Glama introspection check (it has no tools of its own; it gates the upstream's). It belongs in a security-tools list rather than a servers list. Thanks!

PR #13012, opened 2026-08-27
Remove 3 dead links (repos return 404) 🤖🤖🤖

Shyboy0499:
## Summary

Removes 3 entries whose repositories no longer exist (HTTP 404).

| Repo | Status |
| --- | --- |
| `Correctover/mcp-server` | 404 |
| `BlockRunAI/x-grow` | 404 |
| `calclavia/mcp-obsidian` | 404 |

Each was verified with `curl -sS -o /dev/null -w '%{http_code}' -L` against `https://github.com/<repo>`.

## Methodology

A full audit of all ~3,267 GitHub entries found these among ~112 repos that return 404. This PR removes 3 of the clearest cases as a small, verifiable first pass; the rest can be cleaned up in follow-ups if this approach is welcome.

Prepared with AI assistance.


Shyboy0499:
This is a **deletion** PR (3 repos now return 404), not a new server submission — there's no new entry to add a Glama badge to, so the `missing-glama` label is a false positive here.

All three were verified with `curl -L` returning HTTP 404:
- `Correctover/mcp-server`
- `BlockRunAI/x-grow`
- `calclavia/mcp-obsidian`

Happy to adjust the approach if you'd prefer dead-link cleanup handled differently (e.g. batched, or flagged for review rather than removed outright).

PR #13033, opened 2026-08-27
Add smeet666/mcp-imslp to Art & Culture

smeet666:
Adds a read-only MCP server for IMSLP, the Petrucci Music Library.

It searches the works and the people of the library, reads a work with its catalogue numbers, key, instrumentation and dates in the wording the page used, and pages through its editions with the publisher, the editor and the copyright status of each stated per jurisdiction, since a score free in Canada can be protected in the United States.

It downloads no score file: the robots.txt of IMSLP disallows the paths that serve them, so the server builds no address under those paths and links the work page instead. Requests go out one at a time with at least two seconds between them, which is the crawl delay the site publishes.

- npm: https://www.npmjs.com/package/mcp-imslp
- Registry: io.github.smeet666/mcp-imslp
- MIT, TypeScript, stdio, no API key and no account.

Entry placed alphabetically in Art & Culture.


punkpeye:

PR #13135, opened 2026-08-29
Add cogsession to Knowledge & Memory 🤖🤖🤖

premanand8800:
Adds [cogsession](https://github.com/premanand8800/cogsession) under **🧠 Knowledge & Memory**.

**One line added, nothing else touched.**

## What it is

Session memory for AI coding agents. Two things that I think distinguish it from the neighbouring entries:

**An append-only `session.md`** where every entry line carries its own local timestamp, event type and the git state it happened at (`branch@commit+dirty`). That last part is the join back to `git log` — a decision is not a free-floating fact, it was made against a specific state of the code. Because each line is self-describing, a bare `grep` is useful without `-A/-B`:

```
grep -A4 "dead_end"    .cogsessions/*/session.md
grep "main@a1b2c3d"    .cogsessions/*/session.md
```

**Claims that carry their own proof.** You record a statement together with the *command that proves it*. When the files it watches change, the command is re-run:

```
[CogSession] 1 claim(s) no longer hold:
  ✗ the composite key includes the tenant column
    expected '1', got '0'
    asserted in: PR description, line 26
    proof: grep -c 'UNIQUE (a, b, c)' migrations/007_schema.sql
```

No model judgement — it stores a command and re-runs it. Silence means everything still holds.

The motivation, from three weeks of daily use: the expensive failure was never a wrong decision. It was a *right* one that quietly stopped being true. A PR description explaining a schema the code no longer had; a comment naming a constraint that moved; a test asserting a shape the implementation had dropped.

## Checklist

- Category: **Knowledge & Memory**, placed at the top of the section following the recent convention there
- Legend: 🐍 Python · 🏠 local stdio service · 🍎 macOS · 🐧 Linux
- **Windows deliberately not claimed** — the code is pure Python and pathlib and should work, but I have only run it on Linux and macOS, and I would rather under-claim than list a platform I have not verified
- No Glama badge, since the server is not indexed there yet and a broken image helps nobody
- Published and installable: [PyPI](https://pypi.org/project/cogsession/) · listed in the [official MCP Registry](https://registry.modelcontextprotocol.io) as `io.github.premanand8800/cogsession`
- Apache-2.0, 96 tests, CI on 3.11 and 3.13

Title carries `🤖🤖🤖` per CONTRIBUTING.md — this PR was prepared by an agent.


premanand8800:
Thanks — both steps done.

**Listed on Glama:** https://glama.ai/mcp/servers/rea9tn6ajn (`premanand8800/cogsession`, Apache-2.0)

**Badge added** to the entry in the same format as the surrounding lines:

```
[![premanand8800/cogsession MCP server](https://glama.ai/mcp/servers/premanand8800/cogsession/badges/score.svg)](https://glama.ai/mcp/servers/premanand8800/cogsession)
```

The diff is still a single added line.

On the Dockerfile note: there is one at the **repo root**, `https://github.com/premanand8800/cogsession/blob/main/Dockerfile`, so it should be picked up from there. I verified it does what the check needs by sending a real MCP handshake to the running container rather than just confirming it starts:

```
initialize   → protocolVersion 2025-06-18, serverInfo returned
tools/list   → 11 tools
stdout       → pure JSON-RPC, nothing else on it
stderr       → the readiness line, correctly kept off stdout
```

That last point mattered: stdio is the transport, so a stray print to stdout corrupts the stream.

The listing currently shows `tools: []`, so introspection looks like it has not run yet — the crawl only completed within the last day. Happy to wait for it, and if it needs the Dockerfile pasted into the Glama UI rather than read from the repo, say so and I will do that.

premanand8800:
It has been evaluated and the badge is rendering a score.

**Glama listing:** https://glama.ai/mcp/servers/rea9tn6ajn — `premanand8800/cogsession`, Apache-2.0, Python, local.

A build test completed on Glama's side: `01a04ec0-929d-70b9-af88-9ba937e144ad`, v0.1.0, 2026-08-30 00:05.

I also verified independently rather than assuming, by reproducing Glama's own generated Dockerfile verbatim and running its check locally:

```
docker build     succeeds  (debian trixie · node 24 · uv · python 3.14)
container starts "MCP server ready", mcp-proxy serving on 8080
POST /mcp        initialize → 200, serverInfo: cogsession
POST /mcp        tools/list → 11 tools
```

The eleven: `session_init`, `session_checkpoint`, `session_load`, `session_update`, `session_tree`, `session_search`, `session_status`, `session_log`, `claim_record`, `claim_check`, `session_diagram`.

One note in case it saves you time on other PRs: the public directory API currently reports `"tools": []` for this server — but it reports the same for entries already merged here, so it does not appear to be a reliable signal of whether evaluation ran. The badge SVG is the better check. Ours renders with the same three glyphs and the same `#37a169` fill as those merged entries, so it is scoring, not blank.

Happy to adjust anything else. The diff remains a single added line.

PR #10850, opened 2026-07-24
Add arcbounty-mcp (on-chain bounty marketplace for agents) 🤖🤖🤖

Sofiia7:
[arcbounty-mcp](https://github.com/Sofiia7/ARC/tree/main/mcp-server) — Browse, take, and submit USDC bounties on ArcBounty (Arc Network, ERC-8183 escrow + ERC-8004 reputation). Read-only mode needs zero credentials; add a wallet and your agent can earn real USDC.

- Published on npm as `arcbounty-mcp`, also listed in the official MCP Registry (`io.github.Sofiia7/arcbounty-mcp`).
- Added under Finance & Fintech, alongside similar escrow/x402 agent-payment servers already in the list.

PR #13142, opened 2026-08-29
Update Sofiia7/ARC entry: the server now serves Base mainnet too

Sofiia7:
The entry merged in #10850 describes Arc only, which was accurate when it was written. Since then the same server also serves Base mainnet under its own brand, BaseBounty, so the line now names both.

One implementation, one repo: `basebounty-mcp` is a thin wrapper over `arcbounty-mcp` that defaults `ARC_NETWORK=base-mainnet`. Both are published on npm at 0.3.10 and both are in the official MCP Registry.

No new list item, no change to the badge or the emoji flags - only the description text.


Sofiia7:
This is not a duplicate entry - it is an edit of the existing one.

The diff is +1/-1 on the same line: same repo URL, same Glama badge, same 📇 🏠 ☁️ flags. Only the description text changes, because the server now also serves Base mainnet under its own brand (BaseBounty), which the merged description did not mention.

Removing the "duplicate" would mean removing the line I am editing, so there is nothing to strip from the PR. Happy to reword if you would rather the entry stayed Arc-only.

PR #13147, opened 2026-08-29
Update tinqiao-oss/clawtouch-mcp description 🤖🤖🤖

tinqiao-oss:
Maintainer of `tinqiao-oss/clawtouch-mcp` here — refreshing our own entry in **OS Automation**. One line changed, same position in the list.

The existing description covers only the v0.3 tool surface (move / click / drag / type / key combos / scroll). Since then the server also exposes window listing, per-window cropped screenshots, and drawn coordinate markers, which is what lets a vision model point at a target instead of guessing pixels.

The new line also states the platform limits explicitly: the window tooling is Windows-first, macOS needs pyobjc, and Linux has none of it. The old line did not say this, and the 🐧 badge on its own read as though everything worked there.

No other entries or files touched.


tinqiao-oss:
Heads-up on the duplicate check above: this PR does not add a new entry — it **edits the existing** `tinqiao-oss/clawtouch-mcp` line in OS Automation, which is why the bot matches it against itself.

One line changed, same position in the list, no entry added or removed. The existing description covers only the v0.3 tool surface; the update adds window listing, per-window cropped screenshots and coordinate markers, and states the platform limits (window tooling is Windows-first, macOS needs pyobjc, Linux has none) that the old line left implicit while carrying a 🐧 badge.

PR #13204, opened 2026-08-30
Update @saihm/mcp-server-pro install command to the one that onboards 🤖🤖🤖

SAIHM-Admin:
One-line change to an existing Knowledge & Memory entry.

The listed command was `npx -y @saihm/mcp-server-pro`, which starts a stdio MCP server that needs an endpoint and credentials configured before it does anything. Anyone following the entry as written ends up with an inert server and no obvious next step.

`free-join` is the path that provisions an identity and works on a bare machine with no account and no card, so the entry now lists that instead. Nothing else in the entry changed.


SAIHM-Admin:
Thanks — I think the duplicate-check is a false positive here.

This PR doesn't add an entry. It edits the existing `SAIHM-Admin/saihm-mcp-server-pro` line in place (README.md L2491): the diff is one line removed, one line added. Counting the entry on both branches gives the same answer:

- `main`: 1 entry line, 1 occurrence of the repo URL
- this branch: 1 entry line, 1 occurrence of the repo URL

The bot flags it because the *changed* line necessarily still contains the URL that's already listed, which is unavoidable when updating an existing entry rather than adding a new one. So there's nothing to remove — but happy to resubmit differently if in-place edits are better handled another way.

On the glama-check note: the entry already carries the score badge, and has since it was first listed.

The change itself is just the install command, so the listed one is the command that actually onboards. I've also pushed a small correction since opening this: the description said "no account", which overclaimed — the free tier needs no card and no wallet, but it does involve a one-time sign-in, so it now says "no card, no wallet".

PR #13243, opened 2026-08-30
Add DBConvert Streams (Federated SQL) 🤖🤖🤖

slotix:
Adds **DBConvert Streams (Federated SQL)** to the Databases section, in alphabetical order.

What makes it different from the other database servers on the list: **one query can join across sources.** A live PostgreSQL table against a Parquet file in S3, or a CSV someone sent last week against the table it was exported from — federated in-process by DuckDB, with no copying and no staging step.

- Read-only by construction. There is no tool that updates a row, drops a table or alters a schema — absent rather than disabled — and every tool declares `readOnlyHint`, `destructiveHint: false` and `openWorldHint: true`, so a client can check rather than trust.
- Sources are PostgreSQL, MySQL, S3-compatible buckets and folders of Parquet / CSV / JSON files, mixed freely.
- Connection details are given once at startup and never travel through a tool call, so they stay out of the conversation and the logs.
- Runs from connection strings alone: `docker run -i --rm slotix/stream-mcp postgres://…`, or as a one-click `.mcpb` extension for Claude.

Also listed in the official MCP registry as [`com.dbconvert/streams`](https://registry.modelcontextprotocol.io), with the domain verified by DNS.

Docs: https://streams.dbconvert.com/docs/mcp/standalone


slotix:
Done — the server is listed on Glama and the badge is added to the entry.

https://glama.ai/mcp/servers/slotix/dbconvert-streams-public

The build test passes and the listing is installable: Glama built the image, started the server and read its tool list. Worth mentioning, since it is exactly what these checks are for — their build caught a real bug on our side. The Linux binary in our bundle had been compiled on a host with a newer glibc and would not start on `debian:trixie-slim`, or on Ubuntu 24.04, or on Debian 12. It is fixed in 2.7.2, which is what is listed now, and the binary is built in a container from here on.

One thing that may look odd on the listing: the introspection shows **two** tools rather than the full surface. That is intended. The server takes its sources as connection strings at startup, and with none given it exposes only the two core tools; attaching a database, a bucket or a folder adds their groups. We do not bake credentials into an image to inflate that number.

PR #13246, opened 2026-08-30
Add AKzar1el/god-prompt-mcp to Developer Tools 🤖🤖🤖

AKzar1el:
Add AKzar1el/god-prompt-mcp to Developer Tools with its evaluated Glama score badge and supported local platforms. One README entry only.


AKzar1el:
image

PR #13288, opened 2026-08-31
Add gx3-cli-mcp (Mitsubishi MELSEC / GX Works3 ladder analysis) 🤖🤖🤖

purinzan:
Adds [purinzan/gx3-cli-mcp](https://github.com/purinzan/gx3-cli-mcp) under **Industrial & IoT**, placed alphabetically after `FoundryNet/forge-mcp`.

It gives an agent read-only access to Mitsubishi Electric MELSEC PLC projects saved by GX Works3 (`.gx3`) — the programs that run a large share of factory automation in Japan and elsewhere. The agent can trace why a coil never turns on through the ladder logic, find where a device is written and read, search by device comment rather than device number, and separate the conditions that arrive from physical inputs, HMI or network communication (which is how an investigation knows to stop looking at the program and go look at the panel).

Everything runs locally against a project file on the user's own machine. It never writes back to the project, and no project-mutating command is exposed over MCP at all.

- Python, local, Windows-first (CI also runs on Linux and macOS)
- `pip install gx3-cli-mcp`, console scripts `gx3-cli` and `gx3-mcp-server`
- Ships a synthetic project generator, so it can be evaluated without a real customer project
- Unofficial and independent; not affiliated with or endorsed by Mitsubishi Electric

Disclosure per the contributing guide: this PR was prepared by an automated agent (hence the 🤖🤖🤖 opt-in), acting for the repository's author.


punkpeye:
Fixed the entry formatting — moved the badge to the correct position.

PR #13365, opened 2026-09-01
Refresh Pronounce MCP server metadata 🤖🤖🤖

anzy-renlab-ai:
## Summary

Refresh the existing Pronounce entry after continued development:

- update the dictionary size from 1,848 to 1,900+ entries
- add the zero-dependency Node implementation to the language legend
- document the playable MP3 URL returned with pronunciation records
- keep the existing PyPI install and three-tool interface unchanged

## Validation

- confirmed `pronounce-mcp` 0.1.2 is live on PyPI
- confirmed exactly one Pronounce entry remains in the README
- `git diff --check` passes


anzy-renlab-ai:
This PR updates the single existing Pronounce entry rather than adding another listing. The final README contains exactly one `anzy-renlab-ai/pronounce` entry; the diff replaces that line to refresh the entry count, add the Node implementation icon, and document its playable MP3 output. The existing Glama score badge remains in place.

PR #13368, opened 2026-09-01
Add Celmis 🤖🤖🤖

constantinemakoid:
Adds Celmis at the top of Developer Tools.

It is not a wrapper around one API. Celmis indexes a set of repositories into a
symbol graph with tree-sitter — deterministic parsing, no model involved in
building it — and every tool is a different reading of that one index, under one
per-repository access model.

The reason it is worth a line on this list: the index is not a chat log. It is
kept current against the code, it is shared, and it is the same index a
teammate's agent reads. An agent that starts on this codebase tomorrow starts
from what the index knows, not from nothing. To be exact about the boundary —
this is not agent memory and the tools do not write a session's conclusions
anywhere; what persists is the index of the code.

Honest constraint, up front: this is an HTTP mount inside the `api` service of a
`docker compose` stack (Postgres and Qdrant bundled). It is not a standalone
stdio binary you can `npx`. Anyone who already runs Celmis points a client at it;
anyone who does not has to bring the stack up first. Said in the entry itself.

There are already several tree-sitter code indexes on this list and I do not want to
pretend otherwise, so the entry leads on the thing that is actually different rather
than on the parsing: the index has more than one reader and one access model over all
of them. A repository the caller may not research is omitted and named in
`blocked_repos`, not silently dropped.

AGPL-3.0. Docs: https://celmis-labs.github.io/mcp/
The loop this sits inside, 34 seconds: https://youtu.be/jwgbt8vM7XQ
Disclosure: I maintain it.


punkpeye:
Thanks for the PR! The entry format looks good, but there's one outstanding item before we can merge:

- **Glama quality score**: The server has been claimed on Glama, but the quality score hasn't been evaluated yet. Please visit [the Glama score page](https://glama.ai/mcp/servers/Celmis-labs/Celmis/score) and complete the evaluation. Any grade is acceptable as long as the score is set.

Once that's done, we'll be all set to merge. Let me know if you have questions!

PR #13408, opened 2026-09-01
Add ios-agent to OS Automation

emazaheri:
Adds [ios-agent](https://github.com/emazaheri/ios-agent), an MCP server that drives a real iPhone or an iOS Simulator through Apple's XCUIAutomation via WebDriverAgent. Around 30 semantic tools: observe, tap, type, set_value, scroll, launch, deep links, alerts, screenshots, and a policy gate.

It is published on PyPI as [`ios-mcp`](https://pypi.org/project/ios-mcp/) and listed in the [official MCP registry](https://registry.modelcontextprotocol.io) as `io.github.emazaheri/ios-agent`, so `uvx ios-mcp` runs it over stdio.

**On the section:** I put it under **OS Automation** as the closest existing fit, since that section covers system-level UI control and input injection. It is a phone rather than a desktop, so happy to move it to Developer Tools or anywhere else you prefer.

A few things behind the description:

- Perception is budget-aware. A screen reaches the model as a compact digest rather than raw accessibility XML: 251 raw nodes to 12 elements on a third-party screen, 37,000 tokens to 329 on a long list.
- Resolution runs server-side through six tiers, so a failed match retries without spending a model turn.
- Actions return the screen they produced, which halves round trips.
- Destructive actions are classified *before* they run and require approval, through MCP elicitation or a signature the caller passes back.

macOS only, since it builds on Xcode and WebDriverAgent. The simulator path needs no code signing; driving a physical device does.

Legend used: 🐍 Python, 🏠 local service, 🍎 macOS.


emazaheri:
Done, thanks. The server is listed on Glama at https://glama.ai/mcp/servers/emazaheri/ios-agent and the score badge is now in the entry.

One note in case it affects how the score reads: this server drives iOS through Xcode, `simctl` and WebDriverAgent, so it is macOS-only and cannot automate anything from inside a Linux container. The Dockerfile in the repository exists so the server can start and answer introspection, which is what the check asks for, and it says so in its first paragraph rather than implying a container anyone should drive a phone from. Verified against the published package: it returns all 30 tools with no device present, because sessions are created lazily.

It is also on PyPI as [`ios-mcp`](https://pypi.org/project/ios-mcp/) and in the [official MCP registry](https://registry.modelcontextprotocol.io) as `io.github.emazaheri/ios-agent`, so `uvx ios-mcp` runs it over stdio.

PR #13440, opened 2026-09-02
Add ReviewGate to Version Control 🤖🤖🤖

Novohudonossor:
Adds ReviewGate to the Version Control section.

**What it is.** An MCP server for code review against a team's own rules — the conventions written in `.reviewgate/config.yml`, not generic best practice. A cheap generator finds problems, an independent judge refutes them against full files, so the noise stays down. Two tools for the agent loop (`get_team_rules`, `review_changes`), plus a pre-push gate and a bot on GitLab/GitHub pull requests. Single binary, no Node or Docker; bring your own model key.

**Why Version Control** and not Developer Tools: this section covers "code analysis, pull request handling", which is exactly the scope — and Developer Tools already holds 451 entries.

Entry placed alphabetically between `raohwork` and `TamiShaks-2`. I left the score badge out: the licence grade reads F because the EULA is proprietary (free, including commercial use), and a red letter beside green ones would mislead more than inform. Happy to add the badge if you prefer consistency.

---

Unrelated, but since you are the one who will see this: two fields on our Glama listing are not picking up changes. The Description saved on the admin Profile tab (a reload confirms it server-side) still shows the previous text on the search cards — while the display name from the very same JSON-LD block did update. And the GitHub org avatar, set over a day ago, still renders as the default identicon at `glama.ai/avatars/ReviewGate.avif`. Manual "Sync Server" picks up the latest commit but refreshes neither.


Novohudonossor:
Badge added — my mistake: I had left it out deliberately (the licence grade reads F because the EULA is proprietary, though free for commercial use), not realising the badge is a listing requirement rather than a style choice. It is now in the entry, in the standard position.

The server is listed on Glama with the Dockerfile added there and checks passing: quality A, maintenance A.

PR #13461, opened 2026-09-02
Add hono-telescope to Developer Tools 🤖🤖🤖

Jubstaaa:
Adds [hono-telescope](https://github.com/jubstaaa/hono-telescope) under **Developer Tools**, alphabetically between `hivemindunit/llmintel-mcp` and `rm-rf-prod/GroundTruth-MCP`.

It is a debugging tool for [Hono](https://hono.dev) applications whose dashboard endpoint is also an MCP server: five read-only tools let an agent read the running application's recent exceptions together with the request that produced them and that request's logs and queries, recent requests filtered by status or duration, one request in full, and the slowest queries.

- Runs locally inside the user's own app process (`🏠`), TypeScript (`📇`)
- Streamable HTTP transport (`2026-07-28`, with `2025-11-25` accepted)
- Read-only: no tool writes or clears telemetry
- Zero runtime dependencies

Published on npm as `hono-telescope`. Live demo: https://hono-telescope.ilkerbalcilar.com/telescope

Title carries the agent marker per CONTRIBUTING — the change was prepared by an automated agent.


Jubstaaa:
Understood — I removed the badge because it 404'd, not realising the listing requirement runs the other way. Getting the server onto Glama now.

One thing that may need your guidance, because this server's shape is unusual for the list: it is not a standalone process. It is a library that mounts inside the user's own Hono application, so `/telescope/mcp` only exists once someone has added two lines to their app. There is no `npx some-server` to launch.

Two ways that maps onto the Glama check, and I am happy to do whichever you prefer:

1. **Hosted endpoint.** There is already a public one, and it answers introspection today:

   ```
   curl -X POST https://hono-telescope.ilkerbalcilar.com/telescope/mcp \
     -H 'content-type: application/json' \
     -d '{"jsonrpc":"2.0","id":1,"method":"tools/list"}'
   ```

   Returns the five tools; `initialize` reports `hono-telescope 1.1.1`. Per your P.S. this may belong under connectors.

2. **Dockerfile.** The repository ships one that builds the bundled example app, which mounts the middleware and the dashboard and therefore exposes `/telescope/mcp` on port 3000. That gives Glama something that starts and responds to introspection, at the cost of the container being an example app rather than the server itself.

Will submit to Glama and add the badge back once it has a score. Still in progress — not abandoned.

PR #13509, opened 2026-09-02
Add reminal to Command Line 🤖🤖🤖

harshalgajjar:
Adds reminal to Command Line.

reminal exposes the terminals you already have running, across every machine you own, to your agents: list sessions, regex-search live scrollback, read a transcript, and send keys into a PTY. Because the target is just a PTY, the other harness needs no integration — so an agent in one CLI can drive a session belonging to a different vendor CLI, on a different machine.

`reminal integrate` registers it with Claude Code, Codex, Cursor, Gemini, OpenCode, Antigravity and Amp.

Single Go binary, AGPL-3.0, macOS/Linux/Windows. Outbound-only and end-to-end encrypted; the relay is in the same repo and self-hostable.

Repo: https://github.com/harshalgajjar/Reminal


harshalgajjar:
Happy to get it listed on Glama — one question about the Dockerfile requirement first, since reminal is an unusual shape for it.

reminal's MCP server exposes the terminals already running on the host: it lists live PTY sessions across the machines you own, searches their scrollback, reads a transcript, and sends keystrokes into them. It also mirrors GUI windows. All of that is host access by definition.

In a container it will start cleanly and answer introspection — `tools/list` returns the full set — but it will report zero sessions and zero windows, because there's no host to see. So a Dockerfile would pass your check while describing something nobody should actually run. I'd rather not add one that misleads people into `docker run`-ing it.

Two options, whichever suits you:

1. I add a Dockerfile purely for verification, clearly commented as being for registry introspection only, and get the Glama badge.
2. If Glama has a path for host-access servers (the same question applies to anything doing screen capture, USB, or local process control), I'll take that instead.

Let me know which you'd prefer and I'll turn it around quickly.

punkpeye:
Thanks for the question! The Dockerfile requirement is there so Glama can spin up your server in a test environment and verify it responds to `tools/list` and `tools/call` (introspection). Even if your server's full functionality depends on host access, you can create a minimal Dockerfile that starts the server in a way that passes those checks — for example, by starting it and having it respond to the requests (even if it can't actually list running terminals inside the container). The Glama team will let you know if anything else is needed.

Once you've submitted your server on Glama and the quality score is set (any grade is fine), please update your PR to include the Glama badge next to the entry. The badge format is: `[Glama](https://glama.ai/mcp/servers/<your-server-id>)`

Let me know if you have any other questions!

PR #13518, opened 2026-09-02
Mark partymola/fitbit-mcp as retired

partymola:
The Fitbit Web API shuts down on 30 September 2026, so this server is retired. Its successor, partymola/google-health-mcp, is already listed in the same section.

The entry keeps its badge and its position. Only the description changes.


partymola:
This modifies an existing entry rather than submitting a new server. The `partymola/fitbit-mcp` row is already listed, and this PR only changes its description to say it is retired, so the duplicate and missing-badge checks are both firing on the row's own URL. The Glama badge is unchanged and still in the line.

PR #13520, opened 2026-09-02
Add oskarengl/google-tasks-mcp to Workplace & Productivity

oskarengl:
Adds an MCP server for Google Tasks, published on npm as `@oskarengl/google-tasks-mcp`.

It talks to the official Google Tasks API over stdio. The design principle is that the tool surface mirrors what a person can do in the Google Tasks app and nothing more, so there is no planner or agenda view layered on top: list, search and read tasks and lists, create, edit, complete, reopen, reorder, move between lists, and delete. Bring your own OAuth client, so no credentials are held by anyone else.

Single line added under Workplace & Productivity, in alphabetical position. MIT licensed.


oskarengl:
Thanks for the quick response, and for the clear listing requirements. I'm withdrawing this one rather than take up review time: there are already several solid Google Tasks servers on the list, and mine doesn't add enough over them to justify another entry. Thanks for maintaining the list.

PR #13556, opened 2026-09-03
Add FillTrust (security questionnaire corpus) to Security 🤖🤖🤖

erwamartin:
Adds one entry to the Security section.

**Server:** https://www.filltrust.com/mcp — remote, streamable-HTTP at `/api/mcp`.

Guidance for answering vendor security questionnaires: four read-only tools (`search_questions`, `get_question`, `list_standards`, `get_filltrust_posture`) over a corpus of 56 questions drawn from CAIQ v4, SIG Lite, VSA Core, ISO 27001 and GDPR.

A few things that make it unusual enough to be worth a line:

- **No API key, no account, no auth.** `curl` it right now.
- **It calls no model.** Every result is published text returned with the URL of the page it came from, so nothing it says is generated at request time.
- Control references are an allowlist checked in CI against the real control titles, because roughly half of them were wrong when they were written from memory.

It is listed in the official MCP registry as `com.filltrust/questions` (domain-verified via DNS).

The primary link is the documentation page rather than a repository, as with the other hosted servers already in the list (e.g. `pricepertoken.com/mcp`, `shadcnstudio.com/mcp`) — the server is hosted, not distributed as a package.

Marking with 🤖🤖🤖 per CONTRIBUTING.md: this PR was prepared by an agent.


erwamartin:
Thanks — the url-check bot is right and I'll withdraw this.

FillTrust's MCP server is a hosted endpoint (`https://www.filltrust.com/api/mcp`) rather than a distributed package, and there is no public repository behind it, so it cannot meet either the GitHub-link requirement or the Glama score badge, which needs a repo with a Dockerfile.

I read the eleven existing non-GitHub entries in the list as precedent. That was my mistake — they predate the check, and the rule as automated is unambiguous.

Following the P.S. in the Glama bot's comment and listing the hosted endpoint under glama.ai/mcp/connectors instead, which is the route meant for servers shaped like this one. Sorry for the noise.

PR #13559, opened 2026-09-03
Update drolosoft/immich-photo-manager listing description 🤖🤖🤖

juanatsap:
Hi, this updates the description of drolosoft/immich-photo-manager in Multimedia Process. The entry was written for 1.0 (21 tools) and the project has moved a lot since then: it is on 2.0.10 now, with 94 tools and 13 skills, it runs live on Immich 2.x and 3.x before every release, it speaks both MCP protocol eras, and it can be installed as a Claude Code plugin, with uvx, or as a Docker image (ghcr.io/drolosoft/immich-photo-manager).

Thanks for keeping the list. Have a good one!


juanatsap:
On the two bot labels: this PR changes the existing entry in place (same line, same position in Multimedia Process, Glama badge kept), so the URL is already in the README and the bot reads it as a duplicate and does not see the badge. It is the description that changes, from 1.0 to 2.0.11.

Thanks!

PR #13652, opened 2026-09-04
Add serpens-mcp to Search & Data Extraction 🤖🤖🤖

ivashhchenko:
Adds [`ivashhchenko/serpens-mcp`](https://github.com/ivashhchenko/serpens-mcp) to **Search & Data Extraction**, in alphabetical order between `ip2whois/mcp-ip2whois` and `isnow890/naver-search-mcp`.

```
- [ivashhchenko/serpens-mcp](https://github.com/ivashhchenko/serpens-mcp) 🎖️ 📇 ☁️ - Google search for agents: organic results, People Also Ask, related searches, and Google's AI Overview with the sources it names. Markdown output by default. Zero dependencies. Free tier, no card.
```

**Legend used:** 🎖️ official (I maintain the API it wraps), 📇 JavaScript, ☁️ cloud service — it talks to a remote API rather than local software.

**What it does:** two tools, `google_search` and `google_ai_overview`. The second returns Google's AI Overview alongside the organic results, which is the part most search MCP servers do not expose. Output is Markdown by default since that is what a model reads without a parsing step.

**Zero dependencies** — one file of protocol, one HTTP client. For a package that holds an API key, having no third-party code in the path seemed worth the small amount of extra work.

Disclosure, since it is relevant: I maintain the underlying API, which is why the entry carries 🎖️. There is a free tier and no card required, so the server is usable without a purchase.

Tagging per CONTRIBUTING — this PR was prepared by an agent, so the title carries the opt-in marker.


ivashhchenko:
Closing this — reopening from the account that maintains the API, so the submission and the project it points at line up. Same change, no action needed here. Sorry for the noise.

PR #13675, opened 2026-09-05
Add AIgateway MCP server 🤖🤖🤖

Rakesh1002:
Adds [AIgateway](https://github.com/Rakesh1002/ai-gateway) to the Aggregators section — an OpenAI-compatible model aggregator (1,000+ models, 85+ labs, every modality) with a hosted Streamable-HTTP MCP server at api.aigateway.sh/mcp. Model-discovery tools (list/search/per-model schemas) plus invocation tools for chat, image gen, transcription, speech.

One-line entry, alphabetical placement, follows the existing format.


Rakesh1002:
Understood on the new listing requirements — we'll get the server listed on Glama with the checks passing and the Dockerfile added there first, then this entry can be reconsidered. Thanks for maintaining the list.

PR #13719, opened 2026-09-05
Add DevMargin (net margin across 14 billing rails) to Finance & Fintech 🤖🤖🤖

peterbuchlak:
Adds DevMargin under **Finance & Fintech**.

Read-only MCP over your own revenue and net margin across 14 billing rails
(Stripe, Paddle, Polar, Lemon Squeezy, Gumroad, Freemius, RevenueCat, Adapty,
App Store Connect, Google Play and more). It subtracts what each product costs
to run, so an agent can answer which product is losing money *after* costs —
which is the number the individual rail dashboards leave out.

A few things stated plainly, since they affect whether a reader wants it:

- **It is a hosted remote server, and it needs a free account and a token.** That
  is in the entry text rather than left to be discovered after install.
- **Read-only.** The token lists a portfolio and cannot change anything.
- The source is not public, so the link goes to the site rather than a repo —
  following the existing `pricepertoken/mcp-server` entry in the same section.
- Listed on the official MCP registry as `app.devmargin/devmargin`.

One line added, no other changes.


punkpeye:

PR #13838, opened 2026-09-07
Add munim: one MCP server holding a session with every client's cloud account 🤖🤖🤖

vishalsg42:
Adds [munim](https://github.com/vishalsg42/munim) to Cloud Platforms, in alphabetical position between `txn2/kubefwd` and `VmLia/books-mcp-server`.

**What it is.** A coding agent can be logged in to one Cloudflare account, one Vercel, one Resend. Connect a second client and the first goes away. Munim holds a session with all of them at once: each client gets its own registration with the provider, its own token and its own prefix in the tool list, so one agent can read across every client and write inside only the one you name.

It forwards each provider's own MCP tools rather than re-modelling them, supplying the right client's credential, so Cloudflare, Vercel, Resend, Supabase, Linear, Notion, Sentry, Netlify, Zoho and others are reachable with no per-provider code. Alongside that it runs deterministic DNS and email checks (SPF, DKIM, DMARC, MX, CAA, certificates) that never call a model.

Python 3.10+, MIT, no Node and no build step. `uv tool install munim` then `claude mcp add munim -- munim-mcp`.

**On the legend markers:** claimed 🐍 🏠 🍎. Not claiming 🐧 or 🪟 because while CI runs the test suite on Linux, a real end-to-end run on either has not been confirmed, and the project's roadmap says so.


vishalsg42:
Done, thanks for the clear instructions.

- Listed on Glama: https://glama.ai/mcp/servers/vishalsg42/munim
- Score badge added to the entry in the format you specified.

On the P.S.: munim has no hosted endpoint, so the connectors listing does not apply. It is a local stdio server by design, because it holds a separate live OAuth session per client on the operator's own machine. Hosting it would mean several different clients' credentials sitting on shared infrastructure, which is the thing the project deliberately does not do.

PR #13894, opened 2026-09-07
Add dbhq-uk/heliograph to Command Line

grinidx:
Adds one entry to **Command Line**, alphabetically placed.

### What it is

An MCP server for running commands on a machine you **cannot SSH into** — air-gapped, behind a bastion you are not on, or blocked by change-control policy rather than capability.

It is deliberately not another SSH wrapper. The section already has several of those, and they all need a connection heliograph assumes does not exist. Instead it publishes a step to a transport the far side already reaches — git, a file share, S3-compatible storage, a relay, or a file carried by hand — and the whole run comes back as a log with **every line timestamped in UTC, whether it passed or failed**.

There is no tunnel, no proxy and nothing held open. Every command runs because somebody with legitimate access chose to start the loop.

### Why it might interest an MCP audience specifically

The gates are on the far side and a tool call cannot move them. A step must declare itself read-only or an action; an action additionally needs `CONFIRM=yes` and a station started with `--allow-actions`. An MCP client can ask. It does not get to answer.

### Checks

- Go, MIT licensed, released binaries for macOS, Windows and Linux on amd64 and arm64
- `heliograph mcp` — same binary, seven tools, no extra install
- Docs at https://heliograph.dbhq.uk/mcp
- Symbols: 🏎️ Go · 🏠 local service · 🍎 🪟 🐧


grinidx:
Done, thanks.

- Listed on Glama and inspectable: https://glama.ai/mcp/servers/dbhq-uk/heliograph
- Score badge added to the entry in the format above

The server is a Go binary rather than a Node package, so the Dockerfile is in the repo root and Glama builds from its own spec (`heliograph mcp`, no environment variables needed - it answers introspection with no configuration at all and reports honestly that no estates are set up yet).

PR #13969, opened 2026-09-08
Add razi-tools MCP server 🤖🤖🤖

razikallayi:
Adds **razi-tools** under Developer Tools.

28 everyday file and text tools in one server � merge/split/compress PDFs, build PDFs from images, extract PDF text, OCR, compress images and video, generate SQL and QR codes, decode JWTs, diff text, generate mock data.

- **No API key, no signup.** Anonymous calls are rate limited; a free key raises the limit.
- **stdio:** `npx -y razi-tools-mcp`
- **Remote:** `https://www.razi.pro/api/mcp` (JSON-RPC 2.0 over POST)
- **Discovery:** https://www.razi.pro/.well-known/mcp.json
- **Docs:** https://www.razi.pro/developer

One thing worth calling out: `tools/list` only advertises tools that genuinely execute server-side. The site has browser-only tools (WASM background removal, WebRTC file transfer, canvas work) and those are deliberately **not** listed, because a tool a model cannot actually run just costs it a turn to find out. Two more were delisted for the same reason � they need a multipart upload, and MCP has no upload channel, so they could never succeed here.

### Also available as five domain-scoped servers

The combined endpoint above is the whole set. An agent that only does document work should not have to read twenty-eight tool descriptions, so the same tools are also served as coherent, separately-listed servers:

| Endpoint | Tools |
| --- | --- |
| `https://www.razi.pro/api/mcp/documents` | PDF merge/split/compress, build PDFs from images, PDF text extraction, OCR |
| `https://www.razi.pro/api/mcp/media` | image and video compression, image splitting, image generation, QR codes |
| `https://www.razi.pro/api/mcp/web` | URL shortening, screenshots, page metadata, site logos, URL encoding |
| `https://www.razi.pro/api/mcp/text` | text, email, blog-outline and SQL generation, humanizing, diffing, mock data |
| `https://www.razi.pro/api/mcp/dev` | Base64, JWT decoding, JSON formatting, percentages |

A `tools/call` for a tool outside its domain is rejected rather than quietly served, and rate limits are shared across every endpoint, so splitting traffic across domains does not raise the quota.

*Updated: corrected the remote URL from the apex host to `www` (the apex 307-redirects, which browser-based MCP clients do not follow), refreshed the tool count, and added the scoped servers.*


razikallayi:
Thanks for the check — you're right, the badge was showing "not listed". I've removed it rather than leave a broken score in the README; the entry doesn't depend on it. Happy to add it back once the server is evaluated on Glama.

razikallayi:
Badge added — it's listed on Glama now: https://glama.ai/mcp/connectors/io.github.razikallayi/razi-tools

One note on the URL, in case the check is looking for the `/mcp/servers/` form: this server is listed as a **connector** (it's a hosted endpoint at `https://razi.pro/api/mcp`, reachable without an API key), so its badge lives under `/mcp/connectors/...`. The `/mcp/servers/...` path returns "not listed" for it, and the same is true of other connector-listed servers already in this file — e.g. `io.github.tide-foundation/raziel` scores an A under `/mcp/connectors/` and shows "not listed" under `/mcp/servers/`.

Happy to switch the format if you'd prefer something else.

razikallayi:
Update: the Glama listing is now **healthy and ownership-verified**, with all 28 tools introspected.

https://glama.ai/mcp/connectors/io.github.razikallayi/razi-tools

The earlier failed health check turned out to be a real bug on my side, not a Glama issue — worth writing down in case it helps someone else hitting the same thing:

- The connector URL was the **apex** `razi.pro`, which 307-redirects to `www`. Node follows that, so the stdio package worked fine and hid the problem, but a browser-based client can't follow a redirect on a preflighted CORS request. Everything now advertises the canonical host.
- The CORS preflight didn't allow `MCP-Protocol-Version` / `Mcp-Session-Id`.
- `initialize` answered a fixed `2024-11-05` instead of negotiating, and a client sending `Accept: text/event-stream` got `application/json` back.

Badge is in the entry and rendering. Happy to adjust anything else.

punkpeye:

PR #13974, opened 2026-09-08
Add agentic-recall to Knowledge & Memory 🤖🤖🤖

dfrancislyondflabc-tech:
Adds one entry to **Knowledge & Memory**.

**[agentic-recall](https://github.com/dfrancislyondflabc-tech/agentic-recall)** — long-term memory for Claude Desktop and Claude Code across sessions, over a folder of plain markdown files you own and can read, diff and commit.

One rule runs through it: **never infer from wording what a record can settle.** It shows up twice.

**In retrieval.** Similarity search always returns *something*, and ranked by score the least-bad match looks like an answer. Here, when nothing clears the bar, the response carries no results, names the query terms that appear nowhere in the corpus, and offers the nearest document as a labelled weak candidate — not as an answer.

**In claims about code.** A memory saying *"I fixed that in `abc123`"* is not a sentence to be classified; it names a SHA, and a SHA either exists or it does not. Point `MEMORY_GIT_REPOS` at your repositories and every SHA-shaped token in a retrieved memory is looked up: whether it is a real commit, whether it landed on the mainline, its date and subject. Measured over 2,319 ingested exchanges, 707 SHA-shaped candidates collapsed to **355 real commits** — half of what looks like a commit isn't one, which is why every token is checked rather than trusted for its shape.

What it deliberately does *not* do is claim the memory is still true. There is no revert check: a commit that landed and was undone the next day still reports as landed. Instead the response says how many commits have landed since the newest memory was written. The gap between what was said and what has happened is measured, not papered over.

- Hybrid retrieval: BM25F + dense vectors + phrase proximity, fused, then gated
- Embedding model runs locally (~33 MB); nothing leaves the machine
- Node 20/22, MIT
- CI green on macOS, Windows and Linux: 275 checks that need no corpus of your own, plus an end-to-end index-and-retrieve run on each platform
- A one-click `.mcpb` bundle for Claude Desktop is attached to the release

Emoji per the legend: 📇 JavaScript · 🏠 local service · 🍎 🪟 🐧 all three platforms, each proven by its own CI job.

Opting into the 🤖🤖🤖 agent fast-track from CONTRIBUTING.


dfrancislyondflabc-tech:
Thanks for merging it.

Discord username: danielfrancislyon

No remote server to list — agentic-recall runs locally against a folder of markdown files on the user's own machine, so there's nothing hosted to point at connectors.

PR #13828, opened 2026-09-07
Add eoylab/jftc-actions (Legal) 🤖🤖🤖

eoylab:
Adds one line to **Legal**, in alphabetical position, with the Glama score badge already included per the listing requirements — the server is on Glama at <https://glama.ai/mcp/servers/eoylab/jftc-actions>.

**[eoylab/jftc-actions](https://github.com/eoylab/jftc-actions)** — enforcement actions by the Japan Fair Trade Commission.

- **361 actions**: 182 cease-and-desist and related orders under the Antimonopoly Act, 179 recommendations under the subcontract act, 2010-04 to 2026-07
- Every record carries the company as published, the case number, the cited provisions both normalised and verbatim, the date, and a **source URL**
- Four read-only tools: `search_actions`, `get_action`, `list_recent`, `stats`
- **Zero dependencies, no network calls at runtime, no API key** — the data ships as JSON in the repo, so `node src/mcp/server.mjs` is the whole command

**On the working-server requirement:** it answers `initialize` and `tools/list` immediately with no configuration, because there is nothing to configure.

**What it deliberately does not do.** It states no verdict — each record restates an action already taken, and there is no "is this conduct lawful" tool. It stores no representatives' names, although every order names one. And surcharge amounts are null on every record with a status field saying why: they are not in the published index tables, and for several years exist only inside an image, so they are declared missing rather than read by OCR and published as fact.

Source: Japan Fair Trade Commission, reused under the [Japanese Public Data License 1.0](https://www.jftc.go.jp/kiyaku/index.html), which permits commercial use with attribution. Code MIT. Browsable at <https://eoylab.github.io/jftc-actions/>.

---

Opting into the agent fast-track per CONTRIBUTING.md.

PR #13437, opened 2026-09-02
Add eoylab/keihyo-cases (Legal) 🤖🤖🤖

eoylab:
Adds one line to **Legal**, in alphabetical position between `edithatogo/fyi-cli` and `gavelin-ai/mcp`.

**[eoylab/keihyo-cases](https://github.com/eoylab/keihyo-cases)** — Japanese advertising-law enforcement actions under 景品表示法 (the Act against Unjustifiable Premiums and Misleading Representations).

- **151 orders** (措置命令 98 / 課徴金納付命令 53), 2021-04 to 2026-08, published by the Consumer Affairs Agency
- Every record carries the company, the cited provisions, the product, the publication date, the agency's own summary, and a **source URL**
- Four read-only tools: `search_cases`, `get_case`, `list_recent`, `stats`
- **Zero dependencies**, no network calls at runtime, no API key — the data ships as JSON in the repo
- Extraction is deterministic (no model), and unparsed releases are kept in `unparsed.json` with reasons rather than guessed at

**It states no verdict.** Each record restates an order that was already issued; there is no "is this ad legal" tool, because that judgement is made case by case and a mechanical pass/fail list would be misused as a safe harbour.

Source: Consumer Affairs Agency, reused under the [Japanese Public Data License 1.0](https://www.caa.go.jp/terms_of_use/) with attribution. Code MIT. Browsable at <https://eoylab.github.io/keihyo-cases/>.

---

Opting into the agent fast-track per CONTRIBUTING.md. Format, alphabetical order and the `📇 🏠` legend symbols follow the existing entries; happy to adjust anything.

PR #13981, opened 2026-09-08
Add eoylab/jftc-actions (Legal)

eoylab:
Re-submitting as **one server per PR**, per the triage note on #13828 — that PR added two, so it was correctly closed. This one adds a single line.

The sibling server (`eoylab/keihyo-cases`) already has its own PR open at #13437, so nothing here overlaps with it.

[eoylab/jftc-actions](https://github.com/eoylab/jftc-actions) — 361 Japan Fair Trade Commission orders and recommendations (Antimonopoly Act cease-and-desist orders, Subcontract Act recommendations), 2010 onward, with a source URL on every record. Four read-only tools over a bundled JSON snapshot: no network calls, no API key, zero dependencies.

On the quality-score requirement: the Glama badge is in the entry, and the artifact it points at is published — `jftc-actions.mcpb` on `v0.1.0`, sha256 recorded in `server.json` and verified against the downloaded file. Also listed on the official MCP Registry as `io.github.eoylab/jftc-actions`. Glama has the repository indexed but had not picked up the release last time I looked, while the sibling server (same layout, same build) scores TDQS A 3.7 — so that looks like a pending re-crawl rather than a packaging problem. Happy to wait if the score needs to land first.

Two things stated in the entry because they change how it can be used: the server **states no verdict of its own** (each record restates an order already issued and published), and surcharge amounts are left `null` because the published index tables carry no amount column — for some years the figures exist only inside images, and reading them by OCR and presenting the result as fact was rejected.


punkpeye:
Thanks for your contribution! The entry looks good in terms of format, emojis, and description. However, there are two things needed before we can merge:

1. **Claim the server on Glama**: Visit https://glama.ai/mcp/servers/eoylab/jftc-actions and claim it.
2. **Ensure a quality score is evaluated**: After claiming, the Glama system will evaluate the server and set a quality score (any grade is acceptable — it just needs to be evaluated, not necessarily perfect).

Once both are done, let me know or just push updates and I can re-check. Happy to help if you have questions!

PR #13995, opened 2026-09-08
Remove openhonest/umbra from Developer Tools

adamzwasserman:
Withdrawing the entry at the author's request. It was added in #11291 and merged yesterday.

Nothing about the list or the review. The project's own maintainer is taking it back out.

Thank you for merging it, and sorry for the churn.


adamzwasserman:
This one removes an entry rather than adding one, so the Glama listing steps above do not apply to it.

The entry was added in #11291 and merged on 7 September. I am the author and I am withdrawing it. The diff is a single deleted line and touches nothing else.

PR #13996, opened 2026-09-08
Add SHAR Production Metadata MCP 🤖🤖🤖

Ares3333333:
Adds SHAR Production Metadata MCP under Multimedia Process.

The public MIT repository contains a read-only local stdio server with one tested tool for validating rights-aware production manifests. The entry includes SHAR Production attribution and https://sharprod.com/ as the publisher homepage.


punkpeye:

PR #10533, opened 2026-07-20
Add inhabitants/sapiens-mcp to Art & Culture

inhabitants:
Adds `sapiens-mcp`, the official MCP server for Sapiens Sintéticos (an AI creative studio: image, video, article, voice, and music generation). Listed on the official MCP registry as `com.sapiensinteticos/sapiens`. Placed alphabetically under Art & Culture. Public repo, MIT; works remote (streamable HTTP) or local via `npx sapiens-mcp`.

PR #14009, opened 2026-09-08
Add inhabitants/sapiens-mcp to Art & Culture

inhabitants:
Re-submission of #10533, which was closed as a remote server.

`sapiens-mcp` is not remote-only. It ships as an npm package from a public repo and runs locally over stdio (`npx sapiens-mcp`) — CONTRIBUTING sends a server to the remote list when it is "remote-only (just a hosted URL, no installable package)", which is not the case here.

The earlier entry led with a hosted endpoint URL, which is likely what made it read as remote. This one drops that entirely and leads with the package.

One line added, alphabetically placed in Art & Culture.


punkpeye:
Thanks for adding your Sapiens MCP server! The entry looks good — GitHub-hosted, correct format, emojis, and Glama badge are all in place.

Before I can merge this, the last remaining item is that the Glama quality score needs to be evaluated. Glama shows the server exists but isn't claimed yet, and the quality score hasn't been set.

Could you:

1. Submit the server at https://glama.ai/mcp/servers
2. Claim it as the owner
3. Make sure a quality score is in place (any grade is acceptable)

Once that's done, I'll merge this right away. Let me know if you need any help with the process.

PR #13862, opened 2026-09-07
Update jeremiahsay/greencalculus-mcp — 12 tools, 16,000+ factors

jeremiahsay:
A one-line correction to my own entry, merged yesterday in #11996. Both numbers were understated:

- **11 tools → 12.** `lookup_factors` (batch lookup) shipped after the entry text was written. Verified against the live server, not against our docs:
  ```
  curl -s -X POST https://mcp.greencalculus.com \
    -H 'Content-Type: application/json' \
    -d '{"jsonrpc":"2.0","id":1,"method":"tools/list"}'
  ```
  returns 12 tools, and `npx -y greencalculus-mcp` lists the same 12.
- **13,000+ factors → 16,000+.** The corpus is 16,673, as stated at https://greencalculus.com/llms.txt.

No change to the link, the badge, the flags, or the position in the list.

PR #14103, opened 2026-09-10
Update GreenCalculus entry: 16,673 factors, 12 tools 🤖🤖🤖

jeremiahsay:
One-line correction to the existing GreenCalculus entry under 📊 Data Platforms. No new entry, no other lines touched.

**Two numbers, both verified against the live service today:**

- Factors `13,000+` → `16,673` — `GET https://api.greencalculus.com/v1/factors` reports `canonical_total: 16673`
- Tools `11` → `12` — `tools/list` on the live server returns 12; `resolve_factor` and `explain_absence` landed after the entry was written

**One removal:** dropped the trailing *"or connect remotely at https://mcp.greencalculus.com"*. #13862 (which carried these same two corrections) was closed under the remote/hosted split, so I took the hosted endpoint to the new list instead — merged as punkpeye/awesome-remote-mcp-servers#133. This line is now purely the npx/stdio package, which is what it should have been.

Position, Glama badge and the ☁️ marker are unchanged.

https://claude.ai/code/session_01A3qGHq8dutqsXt12xhfskb


jeremiahsay:
Both bot comments are expected here: `duplicate-check` fires because this PR edits the existing GreenCalculus line rather than adding a new one (diff is 1 line changed, 0 added), and the Glama score badge in that line already resolves 200 — it's the same badge the entry has carried since #11996.

jeremiahsay:
Closing in favour of #14406, which consolidates this with the other open GreenCalculus PR. The two were editing the same line from opposite ends — one fixed the counts, the other the org path — so they conflicted with each other and neither could apply cleanly. #14406 is a single diff with both, and nothing else touched. Apologies for the churn.

PR #14126, opened 2026-09-10
Update MCP Emails tool count

Albretsen:
MCP Emails now exposes 17 tools (reads and writes were split into separate tools), so this updates the stale count in its entry. Only the number changed, the rest of the line is untouched. Docs: https://mcpemails.com/docs


Albretsen:
This edits the existing Albretsen/MCPEmails entry rather than adding a new one, so the duplicate and missing-glama labels are false positives: the Glama badge is still on the line. The only change is the tool count (11 -> 17), since reads and writes were split into separate tools.

Albretsen:
Both bot flags are false positives for this PR. It edits the existing Albretsen/MCPEmails entry rather than adding a new one: the diff is one line, changing the tool count from 11 to 17. The duplicate check matches the line being edited. The Glama check only scans new entries, so it misses the badge already on this line. The server is listed on Glama and rated A: https://glama.ai/mcp/servers/Albretsen/MCPEmails

PR #14186, opened 2026-09-11
Correct the heliograph entry: two of the five transports have no station side 🤖🤖🤖

grinidx:
I added [dbhq-uk/heliograph](https://github.com/dbhq-uk/heliograph) in #13894 and the entry claims more than the project ships. This narrows it.

The entry said a step could be published to "git, file share, S3-compatible storage, relay or a carried file". Two of those five have a control side and **no station side**: the object store and the bundle (the carried file). The far side cannot pick either of them up, so naming them makes the list promise something a reader cannot do. The project's own docs say so plainly - the [transports table](https://heliograph.dbhq.uk/transports) marks both "none yet" on the station side, and the [air-gapped page](https://heliograph.dbhq.uk/air-gapped) says "the station cannot read it yet".

Now it names the three transports that work end to end today, each proved by a round trip in CI against a real station: git, a file share, and an HTTPS relay.

No other change - same category, same legend, same badge.


grinidx:
Both bots have fired on a one-line edit - the diff is 1 addition and 1 deletion, on the same line.

**duplicate-check**: `dbhq-uk/heliograph` is already listed because this PR corrects the existing entry under Command Line, added in #13894. There is no new entry and no duplicate to remove.

**glama-check**: the entry already carries its Glama score badge, and the badge is in the unchanged part of the line - https://glama.ai/mcp/servers/dbhq-uk/heliograph is live and scored.

The change itself narrows an overclaim I made in #13894. The description listed five transports; two of them, the object store and the bundle, have a control side and no station side, so the far side cannot pick either up. The project's own docs say so - the [transports table](https://heliograph.dbhq.uk/transports) marks both "none yet" for the station side. It now names only the three that work end to end.

PR #14217, opened 2026-09-12
Update nickol-knx-mcp description 🤖🤖🤖

NickoScope:
Refreshes the description of my own entry (`NickoScope/nickol-knx-mcp`) under 🏠 Home Automation. Link, badge and tags unchanged; only the text after `🐍 🏠 -`.

The existing wording dates from the first release. Since then the server also covers KNX Secure posture, topology checks, device decomposition into communication objects and an as-built handover pack, so the line now reflects what it actually does. "No live-bus access" is kept explicit because that is the safety property of this server.

Opened by an automated agent on behalf of the repo owner, hence the 🤖🤖🤖 opt-in from CONTRIBUTING.


NickoScope:
Closing this — the duplicate-check bot is right that the repo is already listed. This PR only rewrote the description of the existing entry, not added a new one, but that is not worth the bot ping-pong. The live entry stays as is.

PR #14221, opened 2026-09-12
Add tokenchit MCP server 🤖🤖🤖

iyashjayesh:
Adds `tokenchit` under **Developer Tools**.

It reads the usage logs Claude Code, Codex and OpenCode already write on the local machine and exposes four tools: total usage with per-agent and per-model breakdown, tokens per calendar day, a year-in-review recap, and agent detection.

Two things that may be relevant to the list:

- **It cannot make a network request.** A test reads every source file in the package and fails if any of them can open a socket — the package has no allowlisted network module at all. Everything it reports is read from disk.
- Every figure is returned with its caveat as a sibling field, because a model handed an "equivalent cost" number on its own will report it to the user as money they spent, and it is list API rates rather than anything paid.

Installable with `npx -y @tokenchit/mcp`, published with a provenance attestation. MIT.

Checked against CONTRIBUTING before opening: public GitHub repo and an installable package (not remote-only), one server, one line, placed in an existing category, description rather than a pitch. Newest entries in Developer Tools sit directly under the heading, so I followed that rather than strict alphabetical — happy to move it if you would rather it were sorted.


iyashjayesh:
Both steps are done.

**Listed on Glama and passing:** https://glama.ai/mcp/servers/iyashjayesh/tokenchit — claimed, build spec configured, build test succeeded, and release `0.8.0` published. The build installs the package from npm (`npm install -g @tokenchit/mcp@latest`) rather than compiling the monorepo, so a green check reflects the artifact users actually install. All four tools are discovered and scoring.

**Badge added** after the server description, in the format requested.

Thanks for the nudge — the listing surfaced two things worth fixing on our side that we have since addressed.

PR #11870, opened 2026-08-10
Add 22 AIops-tools governed MCP servers (consolidates 22 open PRs)

zw008:
Consolidates the 22 still-open one-line PRs from this family (#10695–#10703, #10705–#10708, #10710–#10718) into a single reviewable change, after realising 22 near-identical PRs is a poor thing to ask a maintainer to work through. I'll close those once this is open.

`AIops-tools/Nutanix-AIops` and `AIops-tools/Postgres-AIops` are already listed (thanks!) and are left untouched — this adds the remaining 22.

**What they are.** Each is a governed MCP server for one self-hosted infrastructure platform. What they have in common is the governance layer rather than the API coverage:

- an audit trail that the MCP tools and the CLI **share the same path to**, so there is no unaudited entry point
- `dry_run` on every destructive operation, with the guards running on the preview path too
- `undo` tokens that capture real prior state read from the server before the change
- token/call budget and runaway breakers

**Placement.** Entries go into the existing category sections (Cloud Platforms, Databases, Monitoring, Security) and are sorted alphabetically within each block. `README.md` only, 22 insertions, 0 deletions.

**The tools** — Proxmox, Veeam, K8s, Network (NAPALM + NetBox), TrueNAS, Ceph, Inference (vLLM/Ray), Monitoring (SolarWinds/PRTG/Zabbix), Compliance, AI-Guardian, Endpoint, Fabric (Meraki), Observability (Prometheus/Grafana), Firewall (OPNsense/pfSense), Container-Host (Docker/Portainer), MySQL, Proxy (Traefik/Caddy/HAProxy), Queue (Redis/RabbitMQ), MinIO, XCP-ng, Identity (Keycloak/authentik), CICD (GitLab/Gitea).

All are Apache-2.0/MIT-0, published to PyPI and the MCP Registry, and 19 of the 24 have been verified against live servers rather than mocks (each repo carries a `docs/VERIFICATION.md` saying what was and was not exercised).

PR #14242, opened 2026-09-12
Add AIops-tools/Identity-AIops governed MCP server 🤖🤖🤖

zw008:
Adds one server: [AIops-tools/Identity-AIops](https://github.com/AIops-tools/Identity-AIops), under **identity**.

Python, self-hostable, MIT. Every MCP tool and every CLI command lands an audit row in a local SQLite DB on the same code path, so there is no unaudited entry point; destructive operations capture their before-state and record an inverse where one exists.

One server per PR, per CONTRIBUTING. A previous consolidated PR (#11870) was correctly closed for bundling; this re-submits that work individually, with the tool counts re-checked against each repo and the categorisation corrected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2467,6 +2467,7 @@ Control smart home devices, home network equipment, and automation systems.
 
 Servers that establish who a person or an agent is and what may be known about them.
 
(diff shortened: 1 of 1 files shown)


punkpeye:

PR #14313, opened 2026-09-13
Data Platforms: greencalculus-mcp moved to the greencalculus org

jeremiahsay:
One entry, already in the list (#11996). The repository moved from `jeremiahsay/greencalculus-mcp` to the `greencalculus` organisation; GitHub redirects the old path, but the Glama badge slug does not follow a rename, so the entry now points at the current path and badge (https://glama.ai/mcp/servers/greencalculus/greencalculus-mcp). Re-sorted to its alphabetical position under the new owner name. Tool count 11 → 12 (`explain_absence`), factor count updated.


jeremiahsay:
Closing in favour of #14406, which consolidates this with the other open GreenCalculus PR. The two were editing the same line from opposite ends — one fixed the counts, the other the org path — so they conflicted with each other and neither could apply cleanly. #14406 is a single diff with both, and nothing else touched. Apologies for the churn.

PR #14321, opened 2026-09-13
add fluttersdk artisan and dusk 🤖🤖🤖

anilcancakir:
Adds two Flutter/Dart MCP servers under Developer Tools, alphabetically between `flipt-io` and `flytohub`.

- **fluttersdk/artisan** ([Glama](https://glama.ai/mcp/servers/fluttersdk/artisan)) is a composable Dart CLI framework and stdio MCP server: scaffolding, code generation, transactional plugin installs, hot reload and a REPL.
- **fluttersdk/dusk** ([Glama](https://glama.ai/mcp/servers/fluttersdk/dusk)) drives a live Flutter app over the VM Service: 35 MCP tools for Semantics snapshots, gestures, screenshots and state observation.

Both are already indexed on Glama and carry score badges. Both are MIT and are the official servers for their packages, hence 🎖️.

On the emoji: the legend has no Dart marker, so these carry scope and OS emoji only (🏠 🍎 🪟 🐧). Both run against a local Flutter project rather than a remote API. Happy to adjust if you would prefer a different convention for Dart.

A third server in the same family, `fluttersdk/telescope`, is not included here yet because it is not on Glama at the time of opening. I will add it in a follow-up once it is listed rather than land an entry without a badge.


punkpeye:
Hey, this PR adds multiple servers. Please submit one server per PR so we can review and integrate each one individually. Feel free to open separate PRs — thanks!

PR #14405, opened 2026-09-14
Fix AnythingMCP entry: it is AGPL-3.0, not source-available

keysersoft:
I maintain AnythingMCP. The entry describes it as "source-available", which stopped being true in June 2026 — it is [AGPL-3.0](https://github.com/HelpCode-ai/anythingmcp/blob/main/LICENSE), an OSI-approved licence. That is the kind of thing people filter on, so it is worth correcting.

Same edit refreshes the adapter count (29 → 188, 26 of which need no API key) and adds the response mapping, knowledge graph, SSO and SCIM that shipped since.

The link, the Glama badge and the full emoji run (📇 🏠 ☁️ 🍎 🪟 🐧) are untouched — only the text after the dash changes. One line, one file.


keysersoft:
This isn't a new entry, it's an edit to the AnythingMCP line that's already in the list. The duplicate check is flagging the line I'm modifying.

What changes is the description only. The project relicensed from BSL 1.1 to AGPL-3.0 back in June and the entry still calls it source-available, which is misleading for anyone filtering by licence. The adapter count was also stale (29, now 188).

On the Glama requirement, the badge was already on that line and I kept it, along with the full emoji run. One line, one file, nothing added or removed.

PR #14406, opened 2026-09-14
Update GreenCalculus entry: org path, 16,673 factors, 12 tools 🤖🤖🤖

jeremiahsay:
One diff replacing two of mine that were conflicting with each other.

**Supersedes #14103 and #14313.** Each fixed half of the same line — #14103 the
counts, #14313 the org path — so they conflicted and neither could apply
cleanly. Both are now closed in favour of this. Sorry for the noise.

What changes on the single line:

| | before | after |
|---|---|---|
| repo path | `jeremiahsay/greencalculus-mcp` | `greencalculus/greencalculus-mcp` |
| Glama badge | `…/jeremiahsay/…` | `…/greencalculus/…` |
| factors | 13,000+ | **16,673** |
| tools | 11 | **12** |

The old `jeremiahsay` path still redirects and returns 200, so nothing was
broken — this just points at the canonical org. The tool count was verified
against `tools/list` on the live server today, not read off the README. Also
notes the remote endpoint next to the npx install, since the registry entry
declares both stdio and streamable-http transports.

Scope check: this is an installable server (`npx -y greencalculus-mcp`, on npm),
so it belongs on this list rather than the remote-only one.


jeremiahsay:
Amended the diff: `exact source cell` → `exact source reference`. We retired the phrase "source cell" internally because it was not true of every factor — not all of them carry a cell-level reference — and I had propagated the old wording here. The entry now matches what the API actually returns. No other change.

PR #14422, opened 2026-09-14
Add solidping (self-hosted uptime monitoring) 🤖🤖🤖

fclairamb:
Adds solidping under Monitoring.

It's an open source uptime monitor, AGPL-3.0, single Go binary, with the MCP server built in rather than bolted on as a separate wrapper. The tools cover checks, incidents, results, maintenance windows and status pages.

Repo: https://github.com/fclairamb/solidping
MCP docs: https://solidping.io/docs/features/mcp

I'm the author. It can be self-hosted or used as a hosted service, hence both 🏠 and ☁️.


fclairamb:
Thanks. One thing that may not fit the automated check: solidping's MCP server is not a standalone stdio binary, it is served over HTTP by the app itself at `/api/v1/mcp`, and it requires a token. An anonymous introspection request gets a 401 by design.

The repo does have a Dockerfile that boots the whole server on SQLite with a single `docker run`, so it starts fine, but the MCP endpoint will still ask for auth.

So two questions:

1. Does that qualify for the servers listing, or should the hosted endpoint go under https://glama.ai/mcp/connectors instead, as your P.S. suggests?
2. If it does qualify, is there a way to give the check a token?

Happy to add the score badge once there is a listing to point at. Docs for the MCP part: https://solidping.io/docs/features/mcp

punkpeye:
Thanks for the questions.

1. **HTTP-based MCP servers are fine** — Glama supports both stdio and HTTP transport. The listing requirement is that the server starts and responds to introspection requests (the `initialize` handshake). If authentication is required for the MCP endpoint, that may cause the Glama check to fail because anonymous introspection won't work. You'll want to either configure Glama to include an auth token or adjust the server to allow unauthenticated introspection. This is something you'd handle when submitting to Glama — their documentation or support can help.

2. **Yes, a Dockerfile is required** for the automated checks. Since you already have one, that's covered.

Once your server is submitted to Glama and passes their checks (any quality score is fine as long as it's evaluated), please update this PR to include the Glama badge in the entry.

You can find the full contribution requirements in the [CONTRIBUTING.md](https://github.com/punkpeye/awesome-mcp-servers/blob/main/CONTRIBUTING.md). Happy to help if you have more questions.

PR #14437, opened 2026-09-15
Add oleg-vdv/kepil (Security) 🤖🤖🤖

oleg-vdv:
Adding Kepil under Security. Installable from PyPI and run locally, so it belongs on this list rather than the remote one.

What the server does: seven tools for listing what agents are allowed to do, creating orders, running steps through the gate, asking whether a given action is permitted by the mandate, reading pending confirmations, and verifying the journal.

What it deliberately does not do: confirm an irreversible action. Only a person can, in the panel or in Telegram, and there is a test whose only job is to fail if a confirm tool ever appears in the tool list. An agent approving its own irreversible actions would make the rest decoration.

The journal is append-only with each record carrying the hash of the previous one, and the verifier is a separate implementation in a different language (Python writes, TypeScript verifies) so a log is not checked only by the tool that wrote it.

AGPL-3.0, Python standard library only with CI failing the build on a third-party import, state in plain JSON files.

Icons: 🐍 Python, 🏠 local service. Added as the last entry of the Security section, which is not alphabetically ordered.

Disclosure: I maintain the project, and the PR title carries the 🤖🤖🤖 opt-in from CONTRIBUTING because this pull request was prepared by an agent.


punkpeye:
Thanks for submitting! There are a few items we need to address before this can be merged:

1. **Claim the server on Glama** — Go to https://glama.ai/mcp/servers, find your server, and claim it.
2. **Get the quality score evaluated** — Once claimed, Glama will run checks and provide a quality score. Any grade is fine, but it must be evaluated (not showing "?" for quality).
3. **Add the Glama badge to your entry** — After the score is assigned, add the badge line to your PR entry (e.g., `[MCP](...)`).

Let me know if you have any questions!

oleg-vdv:
Done, all three items.

The server is claimed on Glama and graded: license A, quality B, maintenance B. Quality sat at a dash for a while because the build could not start — their generated Dockerfile installs a uv-managed Python and then tries to install into it, which uv refuses as externally managed. Adding `--break-system-packages` to the build step fixed it; the container now starts and answers introspection with seven tools.

Badge added to the entry, after the repository link and before the language icons, matching the surrounding lines.

oleg-vdv:
Confirmed — the server is evaluated and the badge now renders a grade rather than a dash.

Glama shows license A, quality B, maintenance B, and the badge SVG reports "MCP server rated B on Glama". The build test passes: the container starts and answers introspection with seven tools over mcp-proxy.

For anyone who hits the same wall, the cause was in the generated Dockerfile rather than the server. The template installs a uv-managed Python and then runs a build step against it, and uv refuses that with "externally managed — this Python installation is managed by uv and should not be modified". Adding `--break-system-packages` to the build step resolved it:

```
uv pip install --break-system-packages --python /usr/local/bin/python .
```

Several neighbouring entries in the Security section show an ungraded quality with the same shape, so this may be worth a line in the listing requirements.

The entry has been rebased onto current main and is mergeable, one added line.

PR #14479, opened 2026-09-15
obsideo-mcp: refresh entry, client-side encryption is now the default

Regan-Milne:
Updating my own entry (`Regan-Milne/obsideo-mcp`), which had gone stale.

Two things in it are no longer true:

- It says **"optional client-side encryption"**. Encryption has been the default since 0.5.0; `put` encrypts client-side with AES-256-GCM using a locally generated, user-held key, and `encrypt: false` is the opt-out.
- The tool list it names, `put/get/ls/rm/usage`, predates `verify`, the no-email trial and the billing tools. The server exposes 13 tools now.

Rather than swap in a longer tool list, I replaced it with what the entry is actually for: encryption by default, and `verify`, which challenges each provider holding an object and checks its signature, so possession can be proven without downloading the object.

One line changed, no formatting or ordering changes.

For context: the official MCP registry entry had been serving a July build for two months, which is how this drift went unnoticed. That is republished now, and I have added a version-drift check on my side so the listings and the package cannot silently diverge again.


Regan-Milne:
Separate from the PR, in case it is useful as Glama feedback.

This server's Glama listing has its tool list and TDQS score frozen at an inspection from 22 July (v0.2.0, 3 tools: get, put, usage). The README-derived page body tracks the repo and is current, so the two halves of the page disagree. The server exposes 13 tools now, and Completeness reads 2/5 for "lacks essential operations like delete, list objects", which exist as `rm` and `ls`.

The listing also shows the licence as MIT. It changed to PolyForm-Shield-1.0.0 after 0.1.0.

The registry entry was republished today (io.obsideo/mcp, 0.1.0 to 0.7.1) if a re-inspection keys off that.

PR #14509, opened 2026-09-16
Add junqt to Aggregators 🤖🤖🤖

k2so-dev:
Adds [k2so-dev/junqt](https://github.com/k2so-dev/junqt) to Aggregators.

Self-hosted MCP gateway: aggregates stdio, Docker and remote servers behind one endpoint per client, with API keys or a built-in OAuth authorization server downstream and proactive upstream OAuth refresh. TypeScript, MIT, multi-arch image on GHCR.


k2so-dev:
Junqt is an MCP gateway, not an MCP server — it has no tool surface of its own. A fresh container starts with an empty database and zero upstreams, so tools/list returns nothing until an operator adds servers; there is nothing for Glama to introspect, and the tool set differs per install.
The one MCP server it ships is administrative (/mcp/_admin, off by default, bearer-token only). It can add stdio servers, i.e. run arbitrary commands on the host, so exposing it as a publicly runnable server isn't something I can do.
Happy to close this if the category isn't a fit — or to keep the entry without a badge, as with the other gateways already in Aggregators (1mcp/agent, Aganium/agenium, AgentHotspot).

PR #14650, opened 2026-09-18
Add token_save_mcp (Developer Tools) 🤖🤖🤖

Habartru:
Adds [token_save_mcp](https://github.com/Habartru/token_save_mcp) under **Developer Tools**.

An MCP server that keeps large files out of the agent's context: `bulk_read` sends the files to a cheap worker model and returns only the answer, so the bytes are paid for in the worker's context rather than the agent's.

What's a little different from the other token-saving entries:

- **A hook, not a suggestion.** A `PreToolUse` hook blocks `Read` above a line/byte threshold and redirects to `bulk_read`, so the saving doesn't depend on the agent remembering the tool exists. Targeted reads (offset/limit) pass through, because editing needs exact text. Enforcement is opt-in and reversible, with a softer `warn` mode.
- **Measured, not estimated.** Every answer carries the worker's real token usage from the provider's `usage` field next to the cost of the read it replaced.
- **Provider-agnostic.** Ollama, OpenRouter, DeepSeek, Groq, or a local runtime — any OpenAI-compatible endpoint.

MIT licensed, `pip install token-save-mcp`, CI green on Linux + macOS across Python 3.10-3.13 (95 offline tests + 19 hook tests, no API calls needed to run them).

Single-line addition to Developer Tools; no other changes.


punkpeye:
Thank you for submitting your server and adding the Glama badge! However, the quality score has not yet been evaluated by Glama. According to our requirements, the server must have a quality score set (any grade is acceptable) to be listed. Please ensure your server is fully evaluated on Glama — you may need to re-submit it or contact Glama support to trigger the evaluation. Once the score appears, this PR can be merged.

PR #14700, opened 2026-09-19
Add suhui-organization/ratchet — compiles a least-privilege policy from an agent's real tool surface 🤖🤖🤖

iversonwuwei:
Adds [suhui-organization/ratchet](https://github.com/suhui-organization/ratchet) under **🔒 Security**.

**What it is.** A local, read-only tool that compiles a least-privilege policy out of an agent's real tool surface:

1. `ratchet scan --home ~` — parses the harness's own config files (executes nothing) and reports every MCP server plus how many are unpinned. On the machine I tested: 16 servers, 12 unpinned.
2. `ratchet scan --introspect` — a separate, explicitly requested step that connects to each server and lists its tools (164 tools across 12 servers in my run).
3. `ratchet policy draft` — emits allow / approve / deny, with the literal reason attached to every verdict, and an explicit "capability can't be inferred" bucket instead of guessing.
4. `ratchet deliver` — a folder with the policy, the report and a sha256 manifest; the recipient recomputes the hashes in their own browser, so the conclusion does not depend on trusting the sender.

**Scope / honesty notes** (please read before merging):

- It is **not** a sandbox and does not block calls; it decides how wide the permission should be and leaves a verifiable record. Tools that enforce at runtime are already in this section and this is complementary to them, not a competitor.
- It records tool names, never arguments, and nothing leaves the machine (no telemetry).
- The capability classifier is a name/description heuristic and is wrong in both directions; it keeps the two false positives I found (`resolve-library-id` matched "format", `sequentialthinking` matched "clear") visible rather than hidden behind a score.

**Install:** `curl -fsSL https://github.com/suhui-organization/ratchet/releases/latest/download/install.sh | sh` or `brew install suhui-organization/tap/ratchet`.

**MCP server:** yes — `ratchet mcp` runs as a stdio server (`ratchet_scan`, `ratchet_policy`, `ratchet_check`), published to the official registry as `io.github.iversonwuwei/ratchet`.

Flags: 🏎️ (Go) 🏠 (local) 🍎 🐧 — there is no Windows build yet, so I did not add 🪟.

I'm an automated agent acting on behalf of the repository owner, which is why the title carries the 🤖🤖🤖 opt-in.


punkpeye:
Thanks for the PR! Almost there — we just need one thing before this can be merged:

- Your server has been submitted to Glama, but the quality score hasn't been evaluated yet. Glama needs to assign the quality score (any grade is fine — it just needs to be set). Could you check on the Glama submission page and ensure everything is in order so the evaluation can complete?

Once that's done, please update the entry to include the Glama badge, and we'll be good to go. Let me know if you have any questions!

PR #14719, opened 2026-09-19
Add guarded-infra-mcp to Cloud Platforms 🤖🤖🤖

anurag444:
Adds [guarded-infra-mcp](https://github.com/anurag444/guarded-infra-mcp) under Cloud Platforms.

Read-only Kubernetes and AWS access for AI agents behind a deny-by-default policy gate. Namespace, resource and region allowlists are enforced in a single middleware that fires on every call, including calls to tools that were never registered, so no tool can bypass it. Every decision — allowed or denied — is appended to a JSONL audit trail that is queried with jq.

What makes it a little different from the other gated-infra servers on the list is that the gate is measured rather than asserted. Two eval suites run in CI with no cluster and no AWS account:

- per-call: 100% block rate across 17 attack fixtures, 0% false positives across 10 legitimate ones
- per-trajectory: whole runs graded on tool selection, order, scope and efficiency, reported separately

MIT licensed, 51 tests, Python 3.11, local stdio server.

Checked against CONTRIBUTING: public GitHub repo, installed and run locally, placed in an existing category, format and legend emoji match the surrounding entries (🐍 Python, ☁️/🏠 talks to both a local cluster and remote AWS APIs, 🍎 🐧).


punkpeye:
Hi @anurag444 — just following up on this. To get your server listed, we need the Glama score badge added to your entry after you've submitted it at https://glama.ai/mcp/servers. Once that's done, let me know and I'll take another look. Thanks!

PR #14731, opened 2026-09-20
Add jiawei686/jev-ultrafast-mcp 🤖🤖🤖

jiawei686:
Adds `jiawei686/jev-ultrafast-mcp` to **Browser Automation**.

An MCP server that hands a whole browser task over instead of paying a turn per click: the caller passes a goal and a URL, and a decision model drives the loop server-side.

Three things it does differently from the browser servers already listed here:

- **Targets come from an element table, never from the model.** It does not write a selector, a coordinate, or JS — it picks a `ref` from the controls the page actually has, and refuses rather than guessing.
- **`browser_assert` is code, not a model.** PASS/FAIL comes from a URL, a text match, or an element state, so the thing that decides did not also do the work.
- **A recorded path replays for free.** Zero model calls per replay, which is what suits the repetitive flows a macro is actually for.

Speaks CDP directly to a Chrome you already have: no Playwright, no Selenium, and no screenshots in the loop. Python, MIT licensed, local service.

- PyPI: [`jev-ultrafast-mcp`](https://pypi.org/project/jev-ultrafast-mcp/)
- MCP registry: `io.github.jiawei686/jev-ultrafast-mcp`

Glama badge added, and the server is already evaluated there (score A, 10 tools).


punkpeye:
Thank you for contributing! The entry looks good, but there is one outstanding issue: the server needs to be [claimed on Glama](https://glama.ai/mcp/servers/jiawei686/jev-ultrafast-mcp) by the repository owner and have its quality score evaluated. Once that's done, we can merge this PR.

PR #14763, opened 2026-09-20
Add emetgate to Developer Tools

ugurcl:
Adds [emetgate](https://github.com/emetgate/emetgate) under Developer Tools.

It is a verification gate that sits between a model and the source tree. The model proposes a replacement body for one symbol, with the hash of the body it expects to replace; the proposal is applied to a shadow copy of the repository, typechecked and tested there, and checked against the rules adopted in an append-only ledger. Only then does anything reach disk. A refusal names the rule and the position, and nothing is written.

MCP tools: `emetgate_skeleton`, `emetgate_symbols`, `emetgate_list`, `emetgate_search`, `emetgate_read_file`, `emetgate_read_symbol`, `emetgate_mutate`, `emetgate_try`, `emetgate_try_batch`, `emetgate_scan`.

Written in Zig, so it has no language emoji in the legend; marked 🏠 (local) and 🪟 (Windows), which are its honest limits today. It parses TypeScript and JavaScript.

MIT licensed, with a published Windows binary and a security policy.


punkpeye:
Hello! The server needs to be listed on Glama before it can be added to the list. Please submit it at https://glama.ai/mcp/servers and then update your PR with the Glama score badge. Let me know if you have any questions about the process.

PR #14770, opened 2026-09-21
Add Agent360dk/computerMCP (Coding Agents) 🤖🤖🤖

GLouv:
New server: [Agent360dk/computerMCP](https://github.com/Agent360dk/computerMCP), added to Coding Agents alphabetically, with the Glama score badge (the server is listed and claimed on Glama).

macOS computer use where the guardrails are the point: secure text fields are blacked out in the image buffer before the PNG is written, every write goes through a consent gate, and the append-only log stores a salted fingerprint of typed text rather than the text itself. 28 tools, 12 of them read-only. MIT, no account, no API key.

*Edit: this PR originally also corrected a stale tool count on our existing browser-mcp line. The duplicate check read that edit as a second listing, so I have dropped it here and will send it separately - this diff is now one added line and nothing else.*


punkpeye:
Hey, a few things to address before this can be merged:

- [ ] **Glama scores**: Your server needs to be evaluated by Glama (quality score must be set). Check your current scores at https://glama.ai/mcp/servers/Agent360dk/computerMCP/score

If you need help, feel free to ask here or on [Discord](https://glama.ai/mcp/discord).

GLouv:
Note on the number in my own line, so nobody is surprised.

The entry says 28 tools. That is what the repo builds and documents. `npx @agent360/computer-mcp` still serves 0.1.0 with 12 tools — 0.2.0 is built and tagged and is waiting on a publish step on our side, and the repo README says so plainly.

The list links the repo, so the line is accurate for what it points at. If you would rather the entry waited until npm matches, say so and I will close this and reopen when it does.

punkpeye:
Thanks for flagging the discrepancy. Per the repo rules, the entry links the GitHub repository, so the description accurately reflects what the repo builds and documents — the npm publish timing doesn't block the merge. However, the entry can't be merged until Glama has evaluated the server and set a quality score; right now the score isn't set yet. Once you've got that in place (it updates automatically once Glama evaluates it), this is ready to go. Feel free to ping me when it's done.

GLouv:
Second correction to my own entry, found by an adversarial review of our own claims an hour ago.

The line says "every write goes through a consent gate" and "runs in the background by default". Measured today: those two cannot both hold. In the shipped default the server offers 15 tools, four of them writing, and **refuses all four** - granting one would need a dialog, and a dialog takes the screen. So the default is in practice read-only, and the only way to write in the background is `CMCP_MODE=allow`, which is precisely the mode that does not ask.

The code is honest about this: the refusal says exactly why. Our copy was not. The repo README and the site now state it plainly, and I would rather the list carried the accurate version:

> macOS computer use with the guardrails on: password fields are blacked out in memory before the screenshot is written and the append-only log never stores typed text in clear. 28 tools, 12 read-only. It runs in the background by default, which means it does not act until you choose - dialogs on, or asking off.

Happy to push that as a commit to this branch if you would rather review the final wording.

PR #14295, opened 2026-09-13
Update Agent360dk/browser-mcp: 40 tools, 20 concurrent sessions

GLouv:
Updating our own entry (line 462), added in #10565. Two numbers went stale:

- **34 tools -> 40** (`mcp-server/tools.js` exports 40)
- **10 concurrent sessions -> 20** (the port range 9876-9895 is 20 ports)

Also added the four words that say what the server is actually for. No new entry, no link changes, one line touched.

PR #14771, opened 2026-09-21
Fix stale counts on the existing Agent360dk/browser-mcp line (no new entry) 🤖🤖🤖

GLouv:
This is **not** a new listing. It edits one line that is already in the list (Browser Automation), because the numbers on it have gone stale.

The line says "34 tools, up to 10 concurrent color-coded sessions". Counted at the source today: **40 tools, 20 sessions**.

One line changed, nothing added, nothing removed.

*Note for the duplicate check: it will flag `Agent360dk/browser-mcp` as already listed. That is the point - this PR corrects that entry rather than adding a second one.*


GLouv:
Closing this as a duplicate of my own older #14295, which makes the same one-line correction and is already in your queue. Two PRs from the same person for one line is noise - that was my mistake.

The correction still stands: the existing Agent360dk/browser-mcp line says "34 tools, up to 10 concurrent color-coded sessions"; counted at the source it is 40 tools and 20 sessions. Please take #14295.

PR #14796, opened 2026-09-21
Add Commander Plus 🤖🤖🤖

yashkhou:
Adds Commander Plus, a local-first stdio MCP control-plane core for persistent workspaces, skill routing, and health guardians.

Repository: https://github.com/yashkhou/commander-plus

The project is public, installable, and self-hosted/local-first as required by this list.


punkpeye:
Thanks for submitting Commander Plus! A couple of items still need to be addressed before this can be merged:

1. **Add the Glama badge** to your entry. The badge should sit right after the GitHub link, in the format used by other entries: `[Glama](glama-server-url)`.
2. **Ensure the quality score is evaluated on Glama.** Maintenance and license grades alone aren't enough — the quality score needs to be evaluated and set on your server's Glama page.

Once both are done, reply here and I'll take another look. Thanks!

PR #14829, opened 2026-09-21
Add JulienRabault/icloud-mcp to Communication 🤖🤖🤖

JulienRabault:
Adding my iCloud Mail MCP server to the Communication section.

**https://github.com/JulienRabault/icloud-mcp** — Python, self-hosted, MIT.

It talks IMAP and SMTP directly to iCloud, using only the standard library for networking and MIME parsing. A few things that are specific to it:

- Searches every folder in one call, not just the inbox. Replies get filed away by mail rules, and an inbox-only search misses them.
- Read tools use `SELECT readonly` and `BODY.PEEK`, so nothing is marked as read as a side effect. There is no delete tool.
- `save_draft` writes to the Drafts folder for review; `send_email` is separate and explicit. `move_emails` simulates unless told otherwise.
- Ships MCP resources and prompts, not just tools, plus standalone scripts for cron use without an MCP client.

44 offline tests, no network or credentials needed to run them.

Entry placed alphabetically in Communication. Happy to adjust the wording or the category if it fits better elsewhere.

[JulienRabault/icloud-mcp MCP server](https://glama.ai/mcp/servers/JulienRabault/icloud-mail)


punkpeye:

PR #9550, opened 2026-07-07
Add equity-intel-mcp (Finance & Fintech)

cstamigo-droid:
Adds cstamigo-droid/equity-intel-mcp — institutional-grade equity analysis for any LLM: SEC insider trades, superinvestor holdings (Dataroma), analyst consensus, valuation, options, and a composite ticker verdict on free data. Python, stdio, local. Public repo.

PR #14846, opened 2026-09-22
Add crypto-intel-mcp (Finance & Fintech) 🤖🤖🤖

cstamigo-droid:
Adds `cstamigo-droid/crypto-intel-mcp` to Finance & Fintech.

Crypto market intelligence over CoinGecko's free API: live quote, multi-timeframe
momentum, trending coins, and a composite verdict. No API key of any kind.

- Repo: https://github.com/cstamigo-droid/crypto-intel-mcp
- Glama: rated **A** (tool definitions A, 4 tools) —
  https://glama.ai/mcp/servers/cstamigo-droid/crypto-intel-mcp

This replaces #9549, which I let go stale back when the Glama quality score was
still the blocker. The score has been in place since July; I have also since
pinned `mcp<2` (a clean install was resolving mcp 2.2.0 and failing on import)
and added the MIT LICENSE file the README badge had always pointed at, so the
server actually starts for anyone who installs it from this list.

Verified before submitting: fresh venv from `requirements.txt`, server starts,
all 4 tools listed over a real MCP stdio handshake, and all 4 return live data
from CoinGecko without credentials.

One line added, badge placed right after the repo link per the surrounding
entries, and ☁️ per the Legend note (the server talks to a remote API).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #14847, opened 2026-09-22
Add equity-intel-mcp (Finance & Fintech) 🤖🤖🤖

cstamigo-droid:
Adds `cstamigo-droid/equity-intel-mcp` to Finance & Fintech.

Equity analysis for any LLM from free public sources: SEC Form 4 insider flow,
superinvestor 13F holdings via Dataroma, valuation, options-implied move, and a
composite ticker verdict.

- Repo: https://github.com/cstamigo-droid/equity-intel-mcp
- Glama: rated **A** (tool definitions A, 7 tools) —
  https://glama.ai/mcp/servers/cstamigo-droid/equity-intel-mcp

Six of the seven tools need no credentials at all. The seventh, analyst consensus,
needs a free Finnhub key and degrades with an honest "no signal" when it is absent,
so the entry says so rather than implying everything is keyless.

This replaces #9550, which I let go stale back when the Glama quality score was
still the blocker. The score has been in place since July; I have also since
pinned `mcp<2` (a clean install was resolving mcp 2.2.0 and failing on import),
added the MIT LICENSE file the README badge had always pointed at, and fixed a
credential leak — the Finnhub key travelled in the query string, and `requests`
puts the request URL in its HTTPError message, so a 401 printed the key into the
tool output. It now travels in a header, with redaction as a backstop and
regression tests that fail if the redaction is removed.

Verified before submitting: fresh venv from `requirements.txt`, server starts, all
7 tools listed over a real MCP stdio handshake.

One line added, badge placed right after the repo link per the surrounding entries,
and ☁️ per the Legend note (the server talks to remote APIs).


cstamigo-droid:
@punkpeye Claimed — https://glama.ai/mcp/servers/cstamigo-droid/equity-intel-mcp now shows the verified owner badge, and the claim link is gone when the page is loaded anonymously, so it should read as claimed from your side too. Quality score was already there (A).

The other two servers in this batch are claimed as well, so #14846, #14847 and #14848 are all ready. Happy to rebase whichever ones go second — they each touch the same README section.

PR #9551, opened 2026-07-07
Add market-data-mcp (Finance & Fintech)

cstamigo-droid:
Adds cstamigo-droid/market-data-mcp — live market data: quotes, news, earnings calendar, a watchlist scanner and a composite verdict over Finnhub's free API, with optional Alpaca hooks. Python, stdio, local. Public repo.

PR #14848, opened 2026-09-22
Add market-data-mcp (Finance & Fintech) 🤖🤖🤖

cstamigo-droid:
Adds `cstamigo-droid/market-data-mcp` to Finance & Fintech.

Live market data for agents over Finnhub's free tier: quotes, company and market
news, earnings calendar, a watchlist scanner and a composite verdict, plus an
optional read-only view of an Alpaca paper-trading account.

- Repo: https://github.com/cstamigo-droid/market-data-mcp
- Glama: rated **A** (tool definitions A, 6 tools) —
  https://glama.ai/mcp/servers/cstamigo-droid/market-data-mcp

The entry states that a free Finnhub key is required, because without it every
Finnhub-backed tool degrades to an honest "no data" rather than failing silently.

This replaces #9551, which I let go stale back when the Glama quality score was
still the blocker. The score has been in place since July; I have also since
pinned `mcp<2` (a clean install was resolving mcp 2.2.0 and failing on import),
added the MIT LICENSE file the README badge had always pointed at, and fixed a
credential leak — the Finnhub key travelled in the query string, and `requests`
puts the request URL in its HTTPError message, so a 401 printed the key straight
into the tool output. It now travels in a header and is stripped where it is read,
with redaction as a backstop and regression tests that fail if the redaction is
removed.

One thing worth flagging honestly: the name `market-data-mcp` on PyPI belongs to
an unrelated project by another author. This one installs from source, and the
README now says so explicitly so nobody installs the wrong package from this list.

Verified before submitting: fresh venv from `requirements.txt`, server starts, all
6 tools listed over a real MCP stdio handshake.

One line added, badge placed right after the repo link per the surrounding entries,
and ☁️ per the Legend note (the server talks to remote APIs).


cstamigo-droid:
@punkpeye Claimed — https://glama.ai/mcp/servers/cstamigo-droid/market-data-mcp now shows the verified owner badge, and the claim link is gone when the page is loaded anonymously, so it should read as claimed from your side too. Quality score was already there (A).

The other two servers in this batch are claimed as well, so #14846, #14847 and #14848 are all ready. Happy to rebase whichever ones go second — they each touch the same README section.

PR #14851, opened 2026-09-22
Add rubics-code/loggerhead-mcp 🤖🤖🤖

gregless22:
Adds Loggerhead under Monitoring, in alphabetical order.

Loggerhead is a native macOS app that receives OpenTelemetry logs, metrics, and traces from the software you run on your own Mac and stores them in a DuckDB file on your disk. The MCP server runs inside the app; the repository holds the `loggerhead-mcp` bridge that connects Claude, Cursor, Codex, Windsurf, or any stdio client to it, as an MCP Bundle on each Release. Seven read-only tools, loopback only, nothing leaves the Mac. The app is required and has a free tier.

- Repository: https://github.com/rubics-code/loggerhead-mcp
- Homepage: https://getloggerhead.com
- Registry: `com.getloggerhead/loggerhead`

Opened by an agent on behalf of the maintainer, with the fast-track marker from CONTRIBUTING.md.


gregless22:
Thanks for the quick check. One thing about this server that matters for the Glama step:

`loggerhead-mcp` is the bridge for a native macOS app. The MCP server runs inside that app; the bridge only connects a client to it over loopback, and it exits when the app is not running. There is nothing a container can start on its own, so a Glama Dockerfile check cannot pass for it, and a score badge would point at a listing that fails introspection by design rather than by fault.

Everything else is in place: the bundle is on a public Release, the server is listed in the Official MCP Registry as `com.getloggerhead/loggerhead` 1.2.0, and every tool is read-only and annotated.

If the list can carry a companion-app server without the Glama badge, this PR is ready as it is. If the badge is a hard requirement, say so and we will close it, with no hard feelings.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #14899, opened 2026-09-22
Add bundlebox (Developer Tools) 🤖🤖🤖

blackswanalpha:
One line under Developer Tools.

[bundlebox](https://github.com/blackswanalpha/bundlebox) answers the orientation questions — where a symbol lives, which files a task touches, what the detectors already found — from a parse and a set difference, so the agent is not billed to grep for them. `bb mcp` serves nine stdio tools with zero dependencies: `bb_pinpoint`, `bb_context`, `bb_snapgen`, `bb_findings`, `bb_scan`, `bb_oversight_brief`, `bb_explain`, `bb_tokens_estimate`, `bb_session`.

The numbers in the entry are from a two-arm run on SWE-bench Verified, 100 instances, both arms counting tokens over text on disk and neither calling a model:

| | bare (grep + read top 10) | bundlebox brief |
|---|---|---|
| context per task | 8.7k tok | 2.8k tok |
| gold files in scope | 24.6% (31/126) | 64.3% (81/126) |

Charts, the per-instance spread and the method: https://blackswanalpha.github.io/bundlebox/docs/benchmark/ — including the one chart that does not favour it, 8.9s to assemble the brief against 246ms to grep. Localisation is a file-level score against the dataset's own patch, not a resolve rate.

Install: `npx -y bundlebox mcp`. MIT, Node >= 20, no runtime dependencies.


punkpeye:
Thanks for the PR! It looks like the server isn't listed on Glama yet. Could you please submit it at https://glama.ai/mcp/servers and then update the PR to include the Glama score badge? Once that's done, we can move forward with merging.

PR #14917, opened 2026-09-22
Move instapdown-mcp and xtapdown-mcp to Social Media 🤖🤖🤖

farukkolip:
Both are Instagram and X creator toolkits, but they sit under Communication while the sibling tiktapdown-mcp is already under Social Media. Moved them there, in alphabetical position next to it.

Also linked the companion sites inline, matching the pattern the tiktapdown-mcp entry already uses.


farukkolip:
This is a move, not an addition — the diff is +2/−2 and adds no new entries.

Both servers were listed under Communication while the sibling
`farukkolip/tiktapdown-mcp` was already under Social Media. The two lines are
removed from Communication and re-inserted in Social Media in alphabetical
position next to it, so the net entry count is unchanged.

The duplicate check appears to only look at added lines, not at the matching
removals in the same diff.

PR #14964, opened 2026-09-23
Add gregordadera/AICB — Roslyn C#/.NET context server 🤖🤖🤖

gregordadera:
Adds `gregordadera/AICB` under Developer Tools, in alphabetical order.

A Roslyn-backed MCP server for C#/.NET: callers and blast radius, implementations and overrides, covering tests, DI resolution, side effects, dead code, namespace cycles, live diagnostics. Installs from nuget.org as a .NET tool (`dotnet tool install -g AIContextBuilder`) on Windows, Linux and macOS; also available as a Windows installer.

On scope: the repository is public and the package is installable, so it is not a remote-only server. It is however closed source — the public repo carries documentation, licence and releases, not the sources. Happy to withdraw the entry if that puts it outside what this list wants.

Opened by an agent, hence the 🤖🤖🤖 opt-in from CONTRIBUTING.


gregordadera:
Withdrawing this one — thanks for the clear listing requirements.

The Glama step is not a route we can take: AICB is closed source (the public repository carries documentation, licence and releases, not the sources), and a Glama listing is built around cloning and building the repository. A Dockerfile that installs the published .NET tool from nuget.org does start and answer introspection — we verified that in CI, 55 tools over a real MCP handshake — but going down that path to satisfy a badge requirement is more than this entry is worth.

No complaint about the requirement itself; it is a sensible way to keep a list of 14,000 entries honest. Closing so it does not sit in your queue.

For anyone who finds this later: the server is listed in the official MCP registry as `io.github.gregordadera/aicb`, and on nuget.org as `AIContextBuilder`.

